PR #6, opened 2017-11-24
Use require stanza instead of script[src=""]

marionebl:
* relative require from script-included js did not work before
* fix by using a top-level require stanza instead as per https://github.com/electron-userland/electron-compile/issues/60


TheReincarnator:
Sau-nice

PR #52, opened 2017-12-07
fix: remove useless type assertion

marionebl:
Currently makes tsc panic on "npm start"


marionebl:
Already fixed on master

PR #85, opened 2017-12-19
Chore/commitlint

kotzendekrabbe:


package.json
```diff
@@ -60,19 +62,23 @@
...
-    "@types/react-dom": "^16.0.3",
     "@types/react": "^16.0.0",
+    "@types/react-dom": "^16.0.3",
+    "commitizen": "^2.9.6",
     "concurrently": "^3.5.1",
     "cz-conventional-changelog": "^2.1.0",
     "electron": "^1.7.9",
     "electron-builder": "^19.49.0",
     "electron-devtools-installer": "^2.2.1",
+    "husky": "^0.14.3",
```
LKuechler: Would we not have to add 'git-hook scripts' to the package.json for husky to do anything?

LKuechler: If i understand it correctly now the only git-hook that we have now is the `commitmsg` script. Is that correct?

kotzendekrabbe: yes

TheReincarnator: Still, where does git know from that the commitmsg script should be called (and I'd prefer to have it removed from the scripts as it is no top-level user script).

marionebl: `husky` installs any of the following `scripts` entries as a githook when `npm install` is executed

* https://github.com/typicode/husky/blob/master/HOOKS.md#hooks

PR #100, opened 2017-12-20
 docs(contributing): add contributing.md to help contributers

### LINKED ISSUE #86: Create a contribution documentation
As a developer I want to have a guideline to create a contribution ;)

Seriously would be nice to have such a thing to make it easier for everybody to contribute to this nice project.

kotzendekrabbe:
Please feel free to give as much feedback as possible + look at gramma etc. thx :)

Fix #86

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Welcome contributors to alva
+# Welcome contributors to Alva
 First of all, we'd love for you to contribute to our source code.
 
 # Code of Conduct
@@ -61,10 +61,10 @@ A good first contribution issue could be
 
 ## Issues
 ### Labels
-- *good first issue* - if you are new to *alva* or open source this could be a good first issue
+- *good first issue* - if you are new to *Alva* or open source this could be a good first issue
 - *help wanted* - issue where we need help
-- *type: bug* - issue for an error/ failure of alva
-- *type: feature* - issue for eature request
+- *type: bug* - issue for an error/ failure of Alva
+- *type: feature* - issue for feature request
 - *type: design* - issue for a design todo
 - *type: documentation* - issue for needed, missing or unclear documentation
 - *status: in progress* - issue is currently in progress
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,113 @@
+# Welcome contributors to alva
+First of all, we'd love for you to contribute to our source code.
+
+# Code of Conduct
+We belive in a welcoming and inclusive environment / community, that's why we enforce our Code of Conduct.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,97 @@
...
+- *type: design* - issue for a design todo
+- *type: documentation* - issue for needed, missing or unclear documentation
+- *status: in progress* - issue is currently in progress
+- *status: has PR* - issue has already an open pull request
+- *priority: low* - issue which is a nice to have, like a backlog issue
+- *priority: mid* - issue needed in the near future, but not super urgent
+- *priority: high* - issue must be resolved as soon as possible
+
+### Working on issues
+Issues should be provided as branches starting with "feat/", as pull requests.
```
LKuechler: Would it make more sense to write something like this:
"Pull requests should be provided as branches starting with 'feat/'"

CONTRIBUTING.md
```diff
@@ -0,0 +1,97 @@
...
+   * backlog board: https://github.com/meetalva/alva/projects/3
+
+
+## Setup for contributers
+To add a contribution to Alva (improve its features, fix issues),
+get the source of both the application itself,
+and the kickstart styleguide (designkit):
+
+* Clone this repository [alva](https://github.com/meetalva/alva): git@github.com:meetalva/alva.git
+* Clone [designkit](https://github.com/meetalva/designkit) repository: git@github.com:meetalva/designkit.git
```
LKuechler: I think i should bring more context to this.
The designkit should not really be needed for the development. It is just a example project.

If the contributer already has a lsg that is compatible with alva there is no need to use the designkit.
If you want i can try to rewrite this part

CONTRIBUTING.md
kotzendekrabbe: as below - only copied this from readme.

Yes, it's more clear.

Also, I would prefer to say - branches should start similar to commit conventions: feat, docs, bug, chore etc.
BUT I don't know what you as project team decided :)

CONTRIBUTING.md
```diff
@@ -0,0 +1,97 @@
...
+* Run in the designkit repository:
+```shell
+npm i && npm run build
+```
+* Run in the main (Alva) respository:
```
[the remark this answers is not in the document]
kotzendekrabbe: this part is only for "contributing" not for modify styling components.
So we only need to show how the contributer setup is , and to be honest - i don't know it.
also a copy & paste text ;)

CONTRIBUTING.md
LKuechler: I thing its a good idea. But i also have the feeling that the name of a branch is not that important. After the merge no one will look at the name again :)

kotzendekrabbe:
@LKuechler thx for your changes :)

PR #125, opened 2017-12-21
chore: simplify commit convention tooling

marionebl:
* use @commitlint/travis-cli to avoid shell script
* use @commitlint/prompt to avoid commitizen config

---
* add .travis.yml to `.editorconfig` tab exceptions and to `.prettierignore`, *.yml does not support tabs
* build on node 8 for speed


marionebl:
Landed via https://github.com/meetalva/alva/commit/45990905329be724b6a83f4263d7950509661608

PR #127, opened 2017-12-21
chore: run tslint on Travis CI

marionebl:
* add lint run-script
* execute npm run lint on Travis CI
* fix known linter issues
* disable rules with false positiives
* add prettier to dev dependencies

depends on #125


src/component/container/element_wrapper.tsx
```diff
@@ -46,7 +46,9 @@ export class ElementWrapper extends React.Component<ElementWrapperProps, Element
 		this.setState({
 			highlight: true
 		});
-		this.props.handleDragStart && this.props.handleDragStart(e);
+		if (typeof this.props.handleDragStart === 'function') {
```
LKuechler: Why the exact check on function?
With the typings it is made sure that there can only be a `function` or `undefined` in `handleDragStart`

marionebl: Sure, but:

1. This makes the needed type clearer when reading
2. tslint did not like the convultion into one expression

PR #128, opened 2017-12-21
chore: run prettier when commiting

marionebl:
depends on #127


package.json
```diff
@@ -24,6 +25,13 @@
   "bugs": {
     "url": "https://github.com/meetalva/alva/issues"
   },
+  "lint-staged": {
+    "*.{ts,tsx}": [
+      "prettier --write",
+      "tslint -c tsconfig.json --fix",
+      "git add"
```
LKuechler: What happens with partial commits? Would they not be added completely?

marionebl: That is true for commits that contain file chunks, don't know if this is common practice in the team?

LKuechler: Good question. I just remembered that we had that problem before in a different project.
But i think we can merge it.

Can you just rebase this branch? I am very sure that a lot of the commits are already in the master. This way it more clear what new stuff comes in

marionebl: Done

PR #136, opened 2017-12-22
Fix typos

tbroadley:
PR created using https://github.com/tbroadley/github-spellcheck-cli.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,16 +2,16 @@
 💓 First of all, we'd love for you to contribute to our source code. 💓
 
 ## Code of Conduct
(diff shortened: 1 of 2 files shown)


LKuechler:
Thanks for the PR and you have a very cool project there.
Can it be that it ignores links? I makes sense to ignore them when they are external but if they are linking to a headline internally that was autocorrected it breaks the link.

tbroadley:
Thanks! You're absolutely right: the tool does ignore links, including internal ones. I didn't consider that fixing a typo in a header could break an internal link. I'll change that behaviour in the next version. Sorry for causing you more work.

LKuechler:
@tbroadley You definitely did not cause me more work. :)
Thanks again for the contribution.

PR #140, opened 2017-12-22
docs(readme): add sinnerschrader to readme

tilmx:


README.md
```diff
@@ -140,4 +140,4 @@ Alva follows the [Contributor Covenant Code of Conduct](CODE_OF_CONDUCT.md).
 
 ---
 
-Copyright 2017. Released under the MIT license.
+Copyright 2017. Proudly powered by [SinnerSchrader](https://github.com/sinnerschrader). Released under the MIT license.
```
LKuechler: maybe move it on its own line above the copyright

tilmx: I actually like it in the footer. The section above should be more about contributing.

LKuechler: What i meant was that is should not be in line with the Copyright :)

tilmx:
Just moved it up.

Issue #89, opened 2017-12-19
Disable application zooming


tilmx:
I can pinch-zoom the application interface (at least on macOS). This shouldn’t happen, so we should fix that.

marionebl:
Are issues closed when landing relevant changes to master or when a new release is cut and downloadable @tilmx?

tilmx:
@marionebl at the moment, when the the PR is merged with the master. Do you think it would make more sense when a new release is out?

marionebl:
No, I am fine either way. Askes to ensure this can be closed.

PR #143, opened 2017-12-22
fix(component): prevent app zooming

glebez:
This PR addresses undesired app zoomig, mentioned in #89 

Based on this discussion int he electron repo: https://github.com/electron/electron/issues/3609

Not fully sure if this is the best place in the code to put those configs.
Point me in the right direction if you know better ;)


LKuechler:
Hi Gleb, thanks for contributing. Looks good to me and i also don't know a better place to put it at the moment :)

PR #155, opened 2017-12-22
Feat/add update check to menu

Jumace:
outsourced the auto-updater to own file
added `Check for Updates` to `Alva` menu
added progress bar for download status of update

### DIFF
--- a/src/electron/electron.ts
+++ b/src/electron/electron.ts
@@ -51,12 +51,7 @@ function createWindow(): void {
 				console.warn('An error occurred: ', err);
 			});
 	}
-	checkForUpdates();
-}
-export function sendStatusToWindow(text: string): void {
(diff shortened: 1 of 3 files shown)


src/electron/electron.ts
```diff
@@ -51,50 +51,17 @@ function createWindow(): void {
...
 			});
 	}
-
-	autoUpdater.checkForUpdatesAndNotify().catch(() => {
-		sendStatusToWindow('Error in auto-updater.');
-	});
+	checkForUpdates(win);
 }
-function sendStatusToWindow(text: string): void {
+export function sendStatusToWindow(text: string): void {
```
LKuechler: can be removed

Jumace: done

PR #186, opened 2017-12-30
Feat/ask before update

LKuechler:


src/electron/auto-updater.ts
```diff
@@ -2,28 +2,60 @@ import { BrowserWindow, dialog } from 'electron';
 import { autoUpdater } from 'electron-updater';
 
 // autoUpdater.logger = log;
+let showUpdateNotAvailable = false;
+let window: BrowserWindow;
```
LKuechler: If someone has a better solution to get these two variables in to the autoUpdater listeners that would be great.

Jumace: Didn't found a better solution either, we would need to rewrite the whole file and I don't think that it's currently worth it.

Issue #142, opened 2017-12-22
remove `console.log` from code


Jumace:
When you go through the code you'll see that some wild `console.log` function appear.
We 'gotta catch 'em all' ;)

tilmx:
Good point!

DanielRuf:
@Jumace do you plan to start the work on a PR?

DanielRuf:
Imho opinion it makes sense to just run these when we are in dev mode.

Jumace:
No I'm not, thought it would be a nice issue for a first timer

-------- Original message --------

DanielRuf:
@maximilianUE would this be interesting for you?

Just to make sure. Should we completely remove them or just not execute the console.log statements in production.

Do they provide any help in dev mode?

LKuechler:
@DanielRuf for now it should be enough to just remove them. A better way would be done with this issue: #154

mimshwright:
I'll work on this. 

You might want to consider adding a linter rule that throws a warning for `console` then run your linter as a precommit hook.

TheReincarnator:
Removed all console.logs, replacing some with info, warn, or error, and removing some entirely.

PR #196, opened 2018-01-10
fix(142): no-console-log

mimshwright:
Removed console logs in src folder (also console.warn, etc). In the case where an error was being caught and logged, i'm allowing the error to be thrown

### DIFF
--- a/src/component/app.tsx
+++ b/src/component/app.tsx
@@ -155,7 +155,6 @@ class App extends React.Component<AppProps> {
 		}
 
 		const designkitPath = PathUtils.join(appPath, 'build', 'designkit');
-		/*console.log(`Design kit path is: ${designkitPath}`);*/
 		dialog.showOpenDialog({ properties: ['openDirectory', 'createDirectory'] }, filePaths => {
 			if (filePaths.length <= 0) {
 				return;
--- a/src/component/app.tsx
+++ b/src/component/app.tsx
@@ -155,7 +155,7 @@ class App extends React.Component<AppProps> {
 		}
 
 		const designkitPath = PathUtils.join(appPath, 'build', 'designkit');
-		console.log(`Design kit path is: ${designkitPath}`);
+		/*console.log(`Design kit path is: ${designkitPath}`);*/
 		dialog.showOpenDialog({ properties: ['openDirectory', 'createDirectory'] }, filePaths => {
 			if (filePaths.length <= 0) {
 				return;
--- a/src/electron/electron.html
+++ b/src/electron/electron.html
@@ -4,9 +4,7 @@
 	<script>
 	const { ipcRenderer } = require('electron');
 
-ipcRenderer.on('message', function(event, text) {
-	// console.log('auto-updater message:', text);
-});
+ipcRenderer.on('message', function(event, text) {});
 </script>
 	<div id="app"></div>
 	<script>require('../../build/component/app.js')</script>
--- a/src/electron/electron.html
+++ b/src/electron/electron.html
@@ -2,12 +2,12 @@
 
 <body>
 	<script>
-		const { ipcRenderer } = require('electron');
+	const { ipcRenderer } = require('electron');
 
-		ipcRenderer.on('message', function (event, text) {
-			console.log('auto-updater message:', text);
-		})
-	</script>
+ipcRenderer.on('message', function(event, text) {
+	// console.log('auto-updater message:', text);
+});
+</script>
 	<div id="app"></div>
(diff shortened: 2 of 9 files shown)


mimshwright:
Fixes #142

DanielRuf:
There are multiple issues with this as it does more than just "removing" the console logs. Removing should also mean deleting the affected lines. The other code should be unchanged imo.

src/component/app.tsx
```diff
@@ -155,7 +155,7 @@ class App extends React.Component<AppProps> {
 		}
 
 		const designkitPath = PathUtils.join(appPath, 'build', 'designkit');
-		console.log(`Design kit path is: ${designkitPath}`);
+		/*console.log(`Design kit path is: ${designkitPath}`);*/
```
LKuechler: It would be better to remove the `console.log`s completely

src/lsg/patterns/link/demo.tsx
```diff
@@ -3,36 +3,22 @@ import Link from './index';
 import * as React from 'react';
 import Space, { Size } from '../space';
 
-const clickHandler = (event: React.MouseEvent<HTMLElement>): void => console.log(event);
+const clickHandler = (event: React.MouseEvent<HTMLElement>): void => {}; //console.log(event);
```
LKuechler: In the demo.tsx files it should be ok to have `console.log`s.
These files are only used when working on the livingstyleguide. This was described wrong in the task.

src/store/page/page_element.ts
```diff
@@ -81,7 +81,7 @@ export class PageElement {
 		const element = new PageElement(pattern, false, parent);
 
 		if (!pattern) {
-			console.warn(`Ignoring unknown pattern "${patternPath}"`);
+			// console.warn(`Ignoring unknown pattern "${patternPath}"`);
```
LKuechler: This one we should probably keep until we have a better logging

LKuechler: I would say that in general the `console.warn`s should be keept and only the `console.log`s should be removed.
In all cases where there is a `console.warn` we have something unexpected what we probably also should show the user.

src/store/store.ts
```diff
@@ -103,7 +103,7 @@ export class Store {
 				}
 			}
 		} catch (error) {
-			console.error(`Failed to open last styleguide or page: ${error}`);
+			throw new Error(`Failed to open last styleguide or page: ${error}`);
```
LKuechler: This one should stay as a `console.error`. :)

src/component/app.tsx
mimshwright: OK.

src/lsg/patterns/link/demo.tsx
mimshwright: Ok.

src/store/page/page_element.ts
mimshwright: Makes ssense.

src/store/store.ts
mimshwright: Understandable. I didn't want to leave any caught but unhandled errors where I had removed a console.warn or console.error. But I'll restore all of these.

mimshwright:
removed commented console.logs. Restored .warn() and .errors()

src/electron/electron.html
```diff
@@ -2,12 +2,10 @@
...
 <body>
 	<script>
-		const { ipcRenderer } = require('electron');
+	const { ipcRenderer } = require('electron');
 
-		ipcRenderer.on('message', function (event, text) {
-			console.log('auto-updater message:', text);
-		})
-	</script>
+ipcRenderer.on('message', function(event, text) {});
```
LKuechler: When we don't have a function that does anything this can be removed :)

mimshwright: I wonder if line 5 should be removed too... `const { ipcRenderer } = require('electron');`

src/electron/electron.html
```diff
@@ -2,12 +2,8 @@
 
 <body>
 	<script>
-		const { ipcRenderer } = require('electron');
-
-		ipcRenderer.on('message', function (event, text) {
-			console.log('auto-updater message:', text);
-		})
-	</script>
+	const { ipcRenderer } = require('electron');
```
LKuechler: Sorry for just noticing this now but now the complete script tag can be removed

faselbaum:
Thanks for the input. I'm going to close the PR since we'll stick to console log for a little bit longer.
In the future we might add a logging abstraction layer and switch to the no-console log rule.

PR #197, opened 2018-01-11
fix(142): tslint-no-console-log

mimshwright:
#142 
Modifies tslint rule for console to throw a warning for console.log, .info, etc. 
console.warn() and error() should still work without errors. I wanted to add the ability to have TSlint stop when building for production while warnings are present but this doesn't appear to be possible with that linter. 

If this is not desired, feel free to reject this PR.

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -63,7 +63,7 @@
 		"no-arg": true,
 		"no-bitwise": true,
 		"no-conditional-assignment": true,
-		"no-console": [false],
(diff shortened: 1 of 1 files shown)


TheReincarnator:
Thanks for your input. As you did not reply to my comment for a while, we would now take care of this ourselves. Cheers!

faselbaum:
Thanks for the input. I'm going to close the PR since we'll stick to console log for a little bit longer.
In the future we might add a logging abstraction layer and switch to the no-console log rule.

PR #207, opened 2018-01-17
chore(store-property): include typings for the BooleanProperty value

Palmaswell:

### DIFF
--- a/src/store/pattern/property/boolean_property.ts
+++ b/src/store/pattern/property/boolean_property.ts
@@ -6,6 +6,9 @@ import { Property } from './property';
  * "true" and "false", and the numbers 1 and 0, as well (see coerceValue()).
  * @see Property
  */
+
+export type BooleanPropertyType = boolean | number | string;
+
 export class BooleanProperty extends Property {
 	/**
 	 * Creates a new boolean property.
@@ -19,8 +22,7 @@ export class BooleanProperty extends Property {
 	/**
 	 * @inheritdoc
 	 */
-	// tslint:disable-next-line:no-any
-	public coerceValue(value: any): any {
+	public coerceValue(value: BooleanPropertyType): boolean {
 		return value === true || value === 'true' || value === 1;
 	}
 


src/store/pattern/property/boolean_property.ts
```diff
@@ -19,8 +22,7 @@ export class BooleanProperty extends Property {
 	/**
 	 * @inheritdoc
 	 */
-	// tslint:disable-next-line:no-any
-	public coerceValue(value: any): any {
+	public coerceValue(value: BooleanPropertyType): boolean {
```
LKuechler: i think this could also be undefined and null or am i missing something?

LKuechler: this function is used in `page_element.tsx` in line 343. The value is not validated there so it could be anything.

Palmaswell: In this case, it's not. The BooleanProperty store value only accepts 3 types a boolean, a number or a string. The coerceVaule function returns a boolean.

Please check comment:

/**
 * A boolean property is a property that supports the values true and false only.
 * As designer content value (raw value), the boolean property accepts the strings
 * "true" and "false", and the numbers 1 and 0, as well (see coerceValue()).
 * @see Property
 */

LKuechler: You know that the comment could also be wrong? ;)
And sure the other any's should also be removed but we should be careful about changing the interface of already used functions.

In the `setPropertyValue` function from the `page_element.ts` the value can be any of the property types we currently support (boolean_property, enum_property, number_array_property, ...).

Every thing that gets thrown in to the coerceValue functions is in the end user input from a text-input, select, doc-tag in there code or anything else. Thats why we can not be sure that the values will always be strings, numbers or a direct boolean

Palmaswell: That is right, the comment could be wrong. 😊

Please don't get me wrong but I think that if we are describing the interface of the BooleanProperty class then we should describe what values does the class methods accepts (`coerceVaule`), and this is only relevant for the **BooleanProperty** all other classes has no knowledge about it. Currently we are using the **BooleanProperty**  as an instance of the Typescript parser and nowhere else.

As far as what I understood  (because I am new to the project) currently the `setPropertyValue` function in the `page_element.ts` supports absutely everything, meaning we have no interface description of the accepted values. So there is no real support for (boolean_property, enum_property and so on).

Would love to hear your thougts @LKuechler 

Thanks!

At the end of the day the minor improvement has no repercusions to the `page_element.ts`. Or am I seeing something wron?

TheReincarnator: coerceValue is inherited from its parent. Is it a good style to reduce the types in sub-classes?
Normally, coerceValue should accept any type of object, as it is its task to transform any object into the required type. If the value is not as recommended by the comment, the implementation should fall back (and will) to a default.

PR #212, opened 2018-01-19
feat: add ability to define a custom preview frame html wrapper

LKuechler:


src/store/project/project.ts
```diff
@@ -30,14 +30,20 @@ export class Project {
...
+	 */
+	@MobX.observable private previewFrame: string;
+
 	/**
 	 * Creates a new project.
 	 * @param id The technical (internal) ID of the project.
 	 * @param name The human-friendly name of the project.
 	 */
-	public constructor(id: string, name: string) {
+	public constructor(id: string, name: string, previewFrame: string) {
```
Jumace: `previewFrame` is missing as a `@param` in the constructor description

LKuechler: done

PR #213, opened 2018-01-19
feat(menu): add context menu for page elements

### LINKED ISSUE #111: Context menu for page elements
As a user I want to have the option to right-click on a page element (in the element pane) and have an option to delete it.

Options when right-clicking an page element: 
- [x] Cut element
- [x] Copy element
- [x] Paste element below (as next sibling)
- [x] Paste element inside (as child)
- [ ] Hide element (#152)
- [x] Delete element

We also should think about other usages of a context menu.

LKuechler:
fixes #111


LKuechler:
This PR will complete these tasks from #111:
- [/] Cut element
- [/] Copy element
- [/] Paste element below (as next sibling)
- [/] Paste element inside (as child)
- [/] Delete element

PR #218, opened 2018-01-27
docs(readme): move greenkeeper badge to the bottom

LKuechler:


Jumace:
As soon as we have more badges we should think about to move them on the top of the MD file, afaik that's how most open source projects are handling badges.

PR #219, opened 2018-01-27
test: add basic build step check for travis

LKuechler:


LKuechler:
@marionebl does it make sense to check if the build steps are working for each PR?

PR #220, opened 2018-01-29
Feat/export page as png

LKuechler:


src/electron/menu.ts
```diff
@@ -92,6 +93,53 @@ export function createMenu(store: Store): void {
...
+							r;
+						`;
+
+						webview.executeJavaScript(code, false, webviewSize => {
+							// set the height of the webview tag to the preview body height
+							// This is needed because capturePage can not capture anything that renders
+							// outside the webview area (https://github.com/electron/electron/issues/9845)
+							webview.style.height = webviewSize.pageHeight;
+
+							dialog.showSaveDialog({}, filename => {
```
Jumace: maybe only support files with png extension? (FileFilter or something like that)

PR #234, opened 2018-02-03
chore(dependencies): update and bind dependencies to newest version

LKuechler:


LKuechler:
This way no dependency that introduces a breaking change with a minor or patch will break Alva.
Greenkeeper will remind us to update these dependencies and will also automatically test them against  `build-lsg` and `build`.

I would be interested about other opinions on this topic.

marionebl:
LGTM, but we should consider using a lockfile for this and check if Greenkeeper supports those or switch to renovateapp if it doesn't.

LKuechler:
Greenkeeper supports lockfiles. What benefits would the lockfile have over the current approach?
(Currently the package-lock.json is also ignored. A few versions ago windows and mac generated different package-lock.jsons every time)

LKuechler:
I will also look at renovateapp. Looks very interesting #239

marionebl:
Lockfiles pin dependencies for the entire tree instead of controlling only the first layer

PR #237, opened 2018-02-06
refactor(store): replace PatternParser with StyleguideAnalyzer

faselbaum:
* Patterns are now managed by the Styleguide Class
* Refactor Pattern detection mechanism to allow for different strategies (StyleguideAnalyzer is now responsible for collecting patterns inside a folder)
* Introduce TypescriptReactAnalyzer to detect patterns based on export information from .d.ts files
* Introduce SyntheticAnalyzer to handle creation of synthetic patterns like Text
* Introduce ReactPattern class
* ReactPattern holds typescript type information to analyze pattern properties. This enables type references to be resolved across files and acts as preparation for object properties.
* Introduce synthetic patterns like the text pattern
* React renderer can now handle class style components

Breakages:
* Styleguides need to provide react typings in order for the analyzer to detect patterns. This breaks "Create Styleguide" functionality since the bundeled styleguide does not provide react typings.

This branch solves issue https://github.com/meetalva/alva/issues/206.


LKuechler:
First some comments to the commits.

**fix(store): import & linter errors**
Why did you not just amend this on the commit where the charges where?
And how was it possible for you to not have something linted when committing? Normally the linter and prettier are triggered on every commit. Did Prettier run on your commits?

**fix(store): remove stuff**
Thats not helpful

**fix(store): linter errors**
Same as the first one

All your commits with the todo files can be removed (rebase interactive)

src/component/container/element_list.tsx
```diff
@@ -93,7 +98,7 @@ export class ElementList extends React.Component<ElementListProps> {
 
 		return {
 			label: key,
-			value: pattern.getName(),
+			value: pattern.name,
```
LKuechler: Why not continue using the get function?

src/component/container/pattern_list.tsx
```diff
@@ -66,20 +71,14 @@ export class PatternListContainer extends React.Component<PatternListContainerPr
...
-				childItem.children = this.createItemsFromFolder(child);
-				result.push(childItem);
-			});
-
-			folder.getPatterns().forEach((pattern: Pattern) => {
+		if (styleguide) {
+			styleguide.patterns.forEach((pattern: Pattern) => {
 				result.push({
-					value: pattern.getName(),
+					value: pattern.name,
```
LKuechler: See above comment with get functions

src/component/container/pattern_list.tsx
```diff
@@ -66,20 +71,14 @@ export class PatternListContainer extends React.Component<PatternListContainerPr
...
 		const result: PatternListContainerItemProps[] = [];
-		if (folder) {
-			folder.getChildren().forEach((child: PatternFolder) => {
-				const childItem: PatternListContainerItemProps = { value: child.getName() };
-				childItem.children = this.createItemsFromFolder(child);
-				result.push(childItem);
-			});
-
-			folder.getPatterns().forEach((pattern: Pattern) => {
+		if (styleguide) {
```
LKuechler: The `styleguide` can not be empty. `styleguide` is always set when calling this function (the function is typed like this :))

src/component/container/pattern_list.tsx
LKuechler: This happens a lot. If you change this please change it for all other places as well

src/component/presentation/preview.tsx
```diff
@@ -88,14 +95,20 @@ export class Preview extends React.Component<PreviewProps> {
...
 			// Then, load the pattern factory
-			const patternPath: string = pattern.getAbsolutePath();
-			let patternFactory: React.StatelessComponent = this.patternFactories[patternPath];
+			const reactPattern = pattern as ReactPattern;
+			const patternPath: string = reactPattern.fileInfo.jsFilePath;
+			let patternFactory: React.StatelessComponent | ObjectConstructor = this.patternFactories[
+				reactPattern.id
+			];
 			if (patternFactory == null) {
 				// tslint:disable-line
```
LKuechler: Is this still needed?

src/store/page/page-element.ts
```diff
@@ -150,7 +170,13 @@ export class PageElement {
 		if (json && (json as JsonObject)['_type'] === 'pattern') {
 			return PageElement.fromJsonObject(json as JsonObject, store, this);
 		} else {
-			const element: PageElement = new PageElement(store.getPattern('text'), false, this);
+			const styleguide = store.getStyleguide('synthetic') as Styleguide;
```
LKuechler: `getStyleguide` returns something with the type `Styleguide`. You should not need to typecast the result

src/store/pattern/pattern.ts
```diff
@@ -14,117 +21,55 @@ import { TypeScriptParser } from './parser/typescript-parser';
...
+	public readonly styleguide: Styleguide;
+
+	/** The StylguideAnalyzer the pattern emerged from. */
+	public readonly analyzer: StyleguideAnalyzer;
 
 	/**
 	 * The ID of the pattern (also the folder name within the parent folder).
 	 */
-	protected id: string;
+	public readonly id: string;
```
LKuechler: Why are you changing them to public? Currently we always use get functions

src/store/pattern/pattern.ts
```diff
@@ -14,220 +18,79 @@ import { TypeScriptParser } from './parser/typescript-parser';
...
-
-	/**
-	 * Returns a string representation of this pattern.
-	 * @return The string representation.
-	 */
-	public toString(): string {
-		const path = this.getRelativePath();
-		return `Pattern(id="${this.id}", path="${path}", properties="${this.properties}")`;
-	}
+	public abstract reload(): void;
```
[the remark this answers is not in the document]
faselbaum: yep. it's currently not needed and not being used. we talked about this. i'm going to remove it.

src/store/styleguide/typescript-react-analyzer/index.ts
```diff
@@ -0,0 +1,114 @@
...
+
+function getPatternName(fileInfo: PatternFileInfo, exportInfo: ReactComponentExport): string {
+	const baseName = PathUtils.basename(fileInfo.jsFilePath, '.js');
+	const directoryName = PathUtils.basename(fileInfo.directory);
+	const name = exportInfo.name ? exportInfo.name : baseName !== 'index' ? baseName : directoryName;
+
+	return name;
+}
+
+function walkDirectoriesAndCollectPatterns(directory: string): PatternFileInfo[] {
```
LKuechler: Should this not be moved outside of the `TypescriptReactAnalyzer`?
This function would also be used by every other analyzer or am i missing something?

src/store/styleguide/typescript-react-analyzer/index.ts
```diff
@@ -0,0 +1,114 @@
...
+		});
+
+		return patterns;
+	}
+}
+
+function getPatternName(fileInfo: PatternFileInfo, exportInfo: ReactComponentExport): string {
+	const baseName = PathUtils.basename(fileInfo.jsFilePath, '.js');
+	const directoryName = PathUtils.basename(fileInfo.directory);
+	const name = exportInfo.name ? exportInfo.name : baseName !== 'index' ? baseName : directoryName;
```
LKuechler: This would make it a bit more readable:
`exportInfo.name || baseName !== 'index' ? baseName : directoryName`

src/store/styleguide/typescript-react-analyzer/index.ts
```diff
@@ -0,0 +1,114 @@
...
+
+		if (FileUtils.lstatSync(childPath).isDirectory()) {
+			patterns = patterns.concat(walkDirectoriesAndCollectPatterns(childPath));
+		}
+	});
+
+	return patterns;
+}
+
+function detectPatternsInDirectory(directory: string): PatternFileInfo[] {
```
LKuechler: Same as above. This will probably also be used in every analyzer

LKuechler:
Some more Todos:

- [ ] Icons for patterns need to be loaded
- [ ] Patterns should be split by folder

src/store/styleguide/typescript-react-analyzer/index.ts
faselbaum: I think we should leave it where it is until we know how pattern detection strategies for other libraries or conventions would work. We also wouldn't know the return type of the function since other conventions require different files to make a valid pattern?

Imagine angular has a template file and a js file that don't neccissarily have to be located in the same directory.

Perhaps you already have a solution in mind?

faselbaum:
i restored the icon detection functionality but this needs some serious work if we don't want to make too many assumptions. The problem is that the icons are inside the patterns source folder. we either add an option to link to the source folder manually or we parse the pattern sourcemaps.

for now it's pretty much a static lookup that maps the pattern folder path to outside the *lib* folder

faselbaum:
working on re-implementation of folders / pattern grouping

faselbaum:
Pattern folders are back. There's a bug in the pattern list though that renders empty folders.
@LKuechler and i talked about this and would like to address the issue in a different PR.

src/store/page/page-element.ts
```diff
@@ -72,17 +65,40 @@ export class PageElement {
 	 * @return A new page element object containing the loaded data.
 	 */
 	public static fromJsonObject(
-		json: JsonObject,
+		json: JsonObject | undefined,
```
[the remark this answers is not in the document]
faselbaum: you are correct. i changed it because i encountered a bug once where json was null. i think this happens when no root element is defined in a page. this shouldn't be handeled here.

i'm going to revert it

src/component/presentation/preview.tsx
```diff
@@ -65,13 +65,17 @@ export class Preview extends React.Component<PreviewProps> {
...
+			const pattern = pageElement.getPattern();
+
+			if (!pattern) {
 				return null;
 			}
 
-			const pattern: Pattern = pageElement.getPattern() as Pattern;
-			if (pattern instanceof TextPattern) {
-				return pageElement.getPropertyValue('text');
+			if (pattern.getType() === PatternType.synthetic) {
```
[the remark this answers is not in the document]
faselbaum: this seems like an already common pattern to me inside the code base. especially in the preview.tsx get functions are used multiple times almost anywhere you look. (eg.: line 86 - 89)

src/store/page/page-element.ts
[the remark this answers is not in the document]
faselbaum: done ✅

src/component/presentation/preview.tsx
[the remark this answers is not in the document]
LKuechler: We should just optimise as we change the files but we could already start here :)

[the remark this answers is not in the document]
faselbaum: 👌 roger that 👌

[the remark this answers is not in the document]
faselbaum: nevermind. i used the wrong variables in the wrong places 😅

src/store/pattern/pattern.ts
```diff
@@ -15,219 +19,77 @@ import { TypeScriptParser } from './parser/typescript-parser';
  * e.g. Patternplate.
  */
 export class Pattern {
-	/**
-	 * The parent folder containing the pattern folder.
-	 */
-	protected folder: PatternFolder;
-
-	/**
-	 * The ID of the pattern (also the folder name within the parent folder).
```
LKuechler: Please re-add the documentation that is still valid.
Probably still valid:
- ID
- iconPath
- constructor

src/store/pattern/pattern.ts
```diff
@@ -15,219 +19,77 @@ import { TypeScriptParser } from './parser/typescript-parser';
...
-	public getFolder(): PatternFolder {
-		return this.folder;
+	public getType(): PatternType {
+		return this.analyzer.getPatternType();
 	}
 
 	/**
-	 * Returns the absolute path to the icon of the pattern, if provided by the implementation.
-	 * @return The absolute path to the icon of the pattern.
+	 * The absolute path to the icon of the pattern, if provided by the implementation.
```
LKuechler: maybe follow the comment naming convention that was created by thomas :)

src/store/pattern/pattern.ts
faselbaum: done ✅

src/store/pattern/pattern.ts
```diff
@@ -15,219 +19,77 @@ import { TypeScriptParser } from './parser/typescript-parser';
...
-		return this.id;
+	public setIconPath(iconPath: string | undefined): void {
+		this.iconPath = iconPath;
 	}
 
 	/**
-	 * Returns the human-readable name of the pattern.
+	 * The human-readable name of the pattern.
 	 * In the frontend, to be displayed instead of the ID.
-	 * @return The human-readable name of the pattern.
```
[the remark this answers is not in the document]
Jumace: ✅

src/store/store.ts
```diff
@@ -232,7 +233,7 @@ export class Store {
 			id.splice(0, 1);
 		}
 
-		const foundElement: PageElement = startElement.getChildren()[id[0]];
+		const foundElement: PageElement = (startElement as PageElement).getChildren()[id[0]];
```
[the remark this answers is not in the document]
faselbaum: oh yeah. didn't really know where the error was coming from. thanks for the hint! :)

PR #243, opened 2018-02-08
feat(rename-pages)

Palmaswell:


src/store/store.ts
```diff
@@ -412,6 +428,22 @@ export class Store {
 		(this.projects as IObservableArray<Project>).remove(project);
 	}
 
+	/**
+	 * Renames the name of the pages files and update the names
+	 * @param id The new ID of the page.
+	 */
+
+	public renameYaml(id: string): void {
```
LKuechler: `renamePage` would be a more descriptive name :)

src/component/container/page_list.tsx
```diff
@@ -11,14 +11,109 @@ export interface PageListProps {
...
+				handleClick={(e: React.MouseEvent<HTMLElement>) => {
+					e.preventDefault();
+					this.handlePageClick(this.props.id);
+				}}
+				handleDoubleClick={this.handlePageDoubleClick}
+				handleKeyDown={this.handlePageKeyDown}
+				name={this.props.name}
+				handleBlur={this.handleBlur}
+				placeholder={this.props.name}
+				value={this.pageName ? this.pageName : this.props.name}
```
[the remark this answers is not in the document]
faselbaum: Or `this.pageName || ''`

src/component/container/page_list.tsx
```diff
@@ -11,14 +11,109 @@ export interface PageListProps {
...
+	@MobX.action
+	protected handlePageDoubleClick(): void {
+		this.pageElementEditable = !this.pageElementEditable;
+	}
+
+	@MobX.action
+	protected handlePageKeyDown(e: React.KeyboardEvent<HTMLInputElement>): void {
+		switch (e.key.toString()) {
+			case 'Escape':
+				this.pageElementEditable = false;
```
LKuechler: You should clear `this.pageName`. If you don't do this it will be auto filled for the next edit process

LKuechler: This should happen whenever you set `pageElementEditable` to false

src/component/container/page_list.tsx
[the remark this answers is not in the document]
Palmaswell: This was @tilmx UX wish. The current pageName should always be prefilled and you can edit it.

src/store/store.ts
Palmaswell: @LKuechler  I will do change this. Got it :)

src/component/container/page_list.tsx
```diff
@@ -11,14 +11,109 @@ export interface PageListProps {
...
+	id: string;
+	name: string;
+	pageRef: PageRef;
+	store: Store;
+}
+
+@observer
+export class PageListItem extends React.Component<PageListItemProps> {
+	@MobX.observable protected pageElementEditable: boolean = false;
+	@MobX.observable protected pageName: string;
```
faselbaum: renaming this to `pageNameInputValue` would make it a little more clearer that this value is not actually the page name. just the editable value in the ui that becomes the page name after applying.

src/component/container/page_list.tsx
Palmaswell: I guess @tilmx expectation is not to completely clear the name. But what I can do is set it to the last available name. What do you think?

src/component/container/page_list.tsx
```diff
@@ -34,13 +129,12 @@ export class PageList extends React.Component<PageListProps> {
 				open={this.pageListVisible}
 			>
 				{this.getProjectPages().map((page: PageRef, index) => (
-					<DropdownItem
-						name={page.getName()}
+					<PageListItem
+						id={page.getId()}
```
LKuechler: it would be great if you could change this to `pageID` then it is clear what it is used for :)

src/component/container/page_list.tsx
```diff
@@ -11,14 +11,109 @@ export interface PageListProps {
...
+				name={this.props.name}
+				handleBlur={this.handleBlur}
+				placeholder={this.props.name}
+				value={this.pageName ? this.pageName : this.props.name}
+			/>
+		);
+	}
+
+	protected handlePageClick(id: string): void {
+		this.props.store.openPage(id);
```
faselbaum: you can just use `this.props.id` here. no need to pass it as an argument since you're already only responsible for one item at a time

src/component/container/page_list.tsx
```diff
@@ -11,14 +11,109 @@ export interface PageListProps {
...
+				handleKeyDown={this.handlePageKeyDown}
+				name={this.props.name}
+				handleBlur={this.handleBlur}
+				placeholder={this.props.name}
+				value={this.pageName ? this.pageName : this.props.name}
+			/>
+		);
+	}
+
+	protected handlePageClick(id: string): void {
```
LKuechler: You don't have to call the function with the id parameter. The function can just use `this.props.id` to get it :)

src/component/container/page_list.tsx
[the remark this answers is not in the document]
LKuechler: Ok then we can remove the placeholder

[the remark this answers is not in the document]
Palmaswell: Yes, actually I should do that @LKuechler I will remove the placeholder. It this ok with you?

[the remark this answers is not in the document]
LKuechler: And you could use @faselbaum suggestion: `this.pageName || this.props.name`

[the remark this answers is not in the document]
faselbaum: then we should set `this.pageName = this.props.name` in the constructor and reset it to `this.props.name` on escaping the input field without hitting enter. Then you can just bind `value={this.pageName}`

src/component/container/page_list.tsx
LKuechler: The problem is with this it feels really strange.
I open the edit mode and write something in. I then press escape because i wan't to write something completely different and when i then go back in to the field i have the value from before.

LKuechler: Example:
rename-page-example

src/store/store.ts
Palmaswell: @LKuechler  Done ✅

src/component/container/page_list.tsx
Palmaswell: Done ✅

src/component/container/page_list.tsx
Palmaswell: Done ✅

src/component/container/page_list.tsx
Palmaswell: Done ✅

src/component/container/page_list.tsx
Palmaswell: Done ✅

src/component/container/page_list.tsx
Palmaswell: Hi @LKuechler,

that is the expect UX behavior, please check my example:

kapture 2018-02-12 at 10 43 38

LKuechler: @Palmaswell your gif shows a completely different use case of the ui.
Please follow these steps:
1. Edit a page name
2. Write something in it
3. Cancel the edit process (esc or click next to the fiel)
4. Now open the edit dialog again

faselbaum:
@tilmx What do you think about the behavior described by @LKuechler ?

src/component/container/page_list.tsx
Palmaswell: Step 4: I could clear to the last saved page name. What do you think about that @LKuechler @tilmx?

LKuechler: @Palmaswell 
I don't think that the last saved page name is needed. If `this.pageNameInputValue` is really cleared the `this.props.name` is used as fallback and with that we have the logic that you and @tilmx want. Without having the bug, that unused values are carried in to the next edit.

Palmaswell: @LKuechler done ✅

PR #247, opened 2018-02-15
Feat/rename project

LKuechler:


src/component/container/page_list.tsx
```diff
@@ -11,14 +11,111 @@ export interface PageListProps {
...
+			case 'Enter':
+				if (!this.pageNameInputValue) {
+					this.pageNameInputValue = this.props.name;
+					this.pageElementEditable = false;
+					return;
+				}
+
+				this.renamePage(this.pageNameInputValue);
+				this.pageElementEditable = false;
+				this.pageNameInputValue = this.props.name;
```
Jumace: If I understand this whole feature right this line doesn't make sense as you set the name already over the 'renamePage' function. As this is asynchronous it could even get a bit dangerous.

src/lsg/patterns/dropdown-item/index.tsx
```diff
@@ -1,5 +1,7 @@
 import { Color, colors } from '../colors';
+
```
Jumace: no need for an empty line here

src/component/container/page_list.tsx
Palmaswell: You're right. It should not be there my bad. Just fixed 👍

src/lsg/patterns/dropdown-item/index.tsx
Palmaswell: done ✅

PR #287, opened 2018-03-08
fix(store): normalize pattern ids across platforms

faselbaum:
This fixes pattern detection on windows


TheReincarnator:
@marionebl says split/join is not that wrong, and he convinced me.
The original paths are OS-specific, then joined OS-specific, and then we should split them OS-specific, and join again using the '/' character, as the join is a pattern ID, not a path.
Additionally, the RegEx construction is not perfectly correct. The string is a regex, but PathUtils.sep is not.
So we should split(PathUtils.sep) and join('/').

faselbaum:
@TheReincarnator done

PR #298, opened 2018-03-13
refactor(store): project and page file structure

TheReincarnator:


src/component/app.tsx
```diff
@@ -80,9 +81,10 @@ class App extends React.Component<AppProps> {
 		// Todo: project and page don't update on page change
 		const project = this.props.store.getCurrentProject();
 		const title = `${project && project.getName()}`;
+		const styleguide = this.props.store.getStyleguide() as Styleguide;
```
LKuechler: do we need the type cast?  `getStyleguide` is typed to return `Styleguide`.

src/component/container/pattern-list.tsx
```diff
@@ -62,7 +63,8 @@ export class PatternListContainer extends React.Component<PatternListContainerPr
 	}
 
 	public render(): JSX.Element {
-		const patternRoot = this.props.store.getStyleguide().getPatternRoot() as PatternFolder;
+		const styleguide = this.props.store.getStyleguide() as Styleguide;
```
LKuechler: same as above

src/component/container/pattern-list.tsx
```diff
@@ -62,7 +63,8 @@ export class PatternListContainer extends React.Component<PatternListContainerPr
...
+		const patternRoot = styleguide.getPatternRoot() as PatternFolder;
```
LKuechler: The only other option here is that `getPatternRoot` can return undefined but especially then it is not a good idea to typecast to something else :)

src/component/app.tsx
TheReincarnator: That's no longer true, I changed that. The styleguide may be undefined now, opening a styleguide creates it. On the other hand, the patterns path is never undefined. I think it is more consistent if the styleguide is not set as long as no styleguide is open, and not that a styleguide with an invalid state exists.
However, in this case, styleguide can actually be undefined, so I should check that better.

src/component/container/pattern-list.tsx
TheReincarnator: The pattern root can no longer be undefined. I will this the getter type.

TheReincarnator: Done.

src/component/app.tsx
TheReincarnator: Done.

src/component/container/pattern-list.tsx
TheReincarnator: Done.

src/store/page/page-element.ts
```diff
@@ -157,7 +158,7 @@ export class PageElement {
 		if (json && (json as JsonObject)['_type'] === 'pattern') {
 			return PageElement.fromJsonObject(json as JsonObject, store, this);
 		} else {
-			const styleguide = store.getStyleguide();
+			const styleguide = store.getStyleguide() as Styleguide;
```
LKuechler: This type cast could probably also be removed and replaced with a check if the styleguide is undefined

TheReincarnator: As discussed, we leave it as it is, before in this case, we can ensure the styleguide always exists.

PR #300, opened 2018-03-13
feat(analyzer): move decision for the renderer in to the analyzers

LKuechler:


TheReincarnator:
Hey @LKuechler, looks very good. I had a slightly different idea (using a base Renderer class instead of an imported function), but I am the OO guy, not the functional guy, and from the design perspective, as far as I can see, it make nore difference. So, I'd approve 👍

PR #302, opened 2018-03-14
fix(store): page elements have stable IDs now, fixing UI lists

TheReincarnator:


TheReincarnator:
When opening an old (not migrated) styleguide for the first time, the element highlighting does not work. The cause is wrong UUIDs (maybe created two times). After saving and reloading, everything is fine.

PR #303, opened 2018-03-14
feat(config): add styleguide option for choosing a specific analyzer

LKuechler:


TheReincarnator:
I find everything great, except the file name, as it is not only a configuration, but also the styleguide's projects and pages. What about "alva.yaml"?

Issue #149, opened 2017-12-22
Allow objects as property


tilmx:
More information following soon...

LKuechler:
@Palmaswell is currently working on this

PR #306, opened 2018-03-16
feat object properties

faselbaum:
Closes https://github.com/meetalva/alva/issues/149

Todo

- [x] Display object properties as tree element
- [x] Store support objects and property paths on`PageElement.getPropertyValue()` & `PageElement.setPropertyValue()`

Note: property garbage collection should be addressed in a separate PR / Issue


src/store/page/page-element.ts
```diff
@@ -1,5 +1,7 @@
+import * as deepAssign from 'deep-assign';
 import { JsonArray, JsonObject, JsonValue } from '../json';
 import * as MobX from 'mobx';
+import * as objectPath from 'object-path';
```
TheReincarnator: Can you please change the case to ObjectPath? We have upper-cased all tool-like imports so far (at least in the store).

src/store/styleguide/pattern.ts
```diff
@@ -147,10 +148,49 @@ export class Pattern {
...
 	 * @return The property for the given ID, if it exists.
 	 */
-	public getProperty(id: string): Property | undefined {
-		return this.properties.get(id);
+	public getProperty(id: string, path?: string): Property | undefined {
+		if (!path) {
+			return this.properties.get(id);
+		}
+
+		const objectProperty = this.properties.get(id);
```
TheReincarnator: Duplicate code. Maybe
```
		const property = this.properties.get(id);
		if (!property || !path) {
			return property;
		}
```

src/store/styleguide/pattern.ts
```diff
@@ -147,10 +148,49 @@ export class Pattern {
...
+		return this.followPath(objectProperty, path, (element, name) => {
+			if (element.getType() !== 'object') {
+				return;
+			}
+
+			return (element as ObjectProperty).getProperty(name);
+		});
+	}
+
+	protected followPath<T>(
```
TheReincarnator: Is this method necessary? Why not just
```
path.split('.').forEach(part => {
	if (property) {
		property = (property as ObjectProperty).getProperty(part);
	}
});
```
without the extra method "followPath" (which also is unnecessarily generic).

Another point: Currently the property nesting relies on the current property value (check for element type 'object'). Instead, the property should yield that it is an object property, like "property.getType() == 'object'", so the final solution could be:
```
path.split('.').forEach(part => {
	if (property) {
		if (property.getType() !== 'object') {
			return undefined;
		}
		property = (property as ObjectProperty).getProperty(part);
	}
});
```

src/store/page/page-element.ts
faselbaum: done

src/store/styleguide/pattern.ts
faselbaum: done

src/store/styleguide/pattern.ts
faselbaum: done

PR #318, opened 2018-03-21
feat(preview): add loading animation

### LINKED ISSUE #299: Function for showing a loading animation while Alva or the Preview loads
We need a function that can be called to create a BrowserWindow while something is loading in the application.
The BrowserWindow should load a simple HTML file.
This way we don't have to wait for react or other scripts to initialise.

In this file we can implement the loading animation from issue #115.

It should be possible to close the loading BrowserWindow from outside to be able to control it inside a event listener for a different window.

The function should then be implemented inside `src/electron/electron.ts` to show the loading spinner until the main application is ready.

goprime10:
Closes #299

### DIFF
--- a/src/electron/preview.html
+++ b/src/electron/preview.html
@@ -2,10 +2,99 @@
 
 <head>
 	<title>alva&trade;</title>
+	
+	<style>
+
+		html,body
+		{
+		    margin: 0;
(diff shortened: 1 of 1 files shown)


LKuechler:
This was merged with a different PR

Issue #274, opened 2018-03-08
Support asset properties (image, file)


TheReincarnator:
The asset is associated directly with the page element, like any other page element value, this issue does not include a central asset management.

TheReincarnator:
@markusoelhafen Can you please make up your mind for a UX here? We have three possible values for asset properties:
* Internal (selected, space-embedded) asset: Choose a file, show a preview
* External (URL, referenced) asset: Enter a URL, show a preview
* Empty: Press clear button to empty the property

Maybe edit the `asset-item/index.tsx` directly?

markusoelhafen:
@TheReincarnator / @LKuechler  **Basic** styling has been applied, can you check the code in `asset-item/index.tsx`?
I left the *clear* button in for now, but we need to think about a general ux/ui for optional properties, so we might not need the clear button in the first place.

PR #324, opened 2018-03-27
Asset properties

TheReincarnator:
As discussed, feature complete, except the following aspects:
* The LSG pattern property-items/asset-item needs UX and HTML/CSS. Currently, it's a draft only. Today, there are three options for the asset: External (URL entered manually), internal (choose-file dialog and saved in the Alva space), and empty.
* In case of an internal asset, the data is stored inline in the page YAML. This is an open discussion.

See Design Kit PR as well: https://github.com/meetalva/designkit/pull/26

Issue is #274


markusoelhafen:
@TheReincarnator / @LKuechler: Basic styling has been applied, can you check the code in asset-item/index.tsx?

TheReincarnator:
@markusoelhafen , @LKuechler We should rework the vertical margins some day. IMHO, the property should have margin, not the input. Now that the buttons are the last element, there is no margin at all.

TheReincarnator:
But for now, and to resolve the long list of pending PRs, I'd like to merge this state.

PR #347, opened 2018-04-06
Switch between pages

TheReincarnator:


src/lsg/patterns/icons/index.tsx
```diff
@@ -22,7 +22,7 @@ export interface IconProps {
 
 export enum Size {
 	XXS = 12,
-	XS = 18,
+	XS = 15,
```
Jumace: As I don't know all of the code, is this a visual breaking change? Was this size used beforehand?

markusoelhafen:
As a quickfix I added a fixed width for the title, with text truncation with ellipsis. 

We could use an improved Title bar behaviour as a new issue: There is a nice ux currently used by zeplin, where the arrows don't jump as long as the user hovers the title bar.
As soon as the user unhovers(?) the element, the width fits to the full length of the title. This way it would be easier to quickly switch between multiple pages.

PR #349, opened 2018-04-06
docs: commit message convention

TheReincarnator:


LKuechler:
even better would be for the commit to be something along the lines of `update commit message convention`

TheReincarnator:
@LKuechler Adjusted commit message, please approve again.

PR #364, opened 2018-04-09
docs(contributing): npm-only support

TheReincarnator:


CONTRIBUTING.md
```diff
@@ -42,6 +42,8 @@ git clone git@github.com:meetalva/alva.git
 npm i && npm start
 ```
 
+(Note: Please use *npm* only. We dropped the support for Yarn for now, as we wired our release processed to the package-lock.json file.)
```
tilmx: Shouldn’t it be "process" instead of "processed"?

TheReincarnator: Yes. I fixed that.

PR #365, opened 2018-04-09
(feat): page-preview

Palmaswell:
Implementation of the page preview


src/component/page-list/page-tile-container.tsx
```diff
@@ -0,0 +1,94 @@
...
+		}, 300);
+	}
+
+	protected handleDoubleClick(e: React.MouseEvent<HTMLElement>): void {
+		Store.getInstance().openPage(this.props.page.getId());
+	}
+
+	@MobX.action
+	protected handleKeyDown(e: React.KeyboardEvent<HTMLInputElement>): void {
+		switch (e.key.toString()) {
```
LKuechler: `e.key` should already be a string or am i missing something?

src/component/page-list/page-tile-composite.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+
+import { observer } from 'mobx-react';
+import { PreviewTile, PreviewTileProps } from '../../lsg/patterns/preview-tile';
+import Space, { Size } from '../../lsg/patterns/space';
+
+export const PageTileComposite: React.StatelessComponent<PreviewTileProps> = observer(
```
TheReincarnator: Why is this a React function and not a regular React class?

TheReincarnator: Does it make sense to have a composite component when its only purpose is a spacer? Much boiler-plate code here. Maybe put that composition code into the PreviewTile or container?

src/component/page-list/page-tile-container.tsx
```diff
@@ -0,0 +1,94 @@
...
+	@MobX.action
+	protected handleChange(e: React.ChangeEvent<HTMLInputElement>): void {
+		this.inputValue = e.target.value;
+	}
+
+	@MobX.action
+	protected handleClick(e: React.MouseEvent<HTMLElement>): void {
+		this.props.onClick(e);
+
+		setTimeout(() => {
```
TheReincarnator: Timeout solutions are almost always dirty and incorrect ;-)

src/component/page-list/page-tile-container.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+@observer
+export class PageTileContainer extends React.Component<PageTileContainerProps> {
+	@MobX.observable public editable: boolean = false;
+	@MobX.observable public inputValue: string = '';
+
+	public constructor(props: PageTileContainerProps) {
+		super(props);
+		this.inputValue = this.inputValue ? this.inputValue : this.props.page.getName();
```
TheReincarnator: What is the difference between this an
```
this.inputValue = this.props.page.getName()
```

src/component/page-list/page-tile-container.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+		this.handleChange = this.handleChange.bind(this);
+		this.handleClick = this.handleClick.bind(this);
+		this.handleDoubleClick = this.handleDoubleClick.bind(this);
+		this.handleKeyDown = this.handleKeyDown.bind(this);
+	}
+
+	@MobX.action
+	protected handleBlur(): void {
+		this.editable = false;
```
TheReincarnator: Blur should commit the change, shouldn't it?

src/component/container/app.tsx
```diff
@@ -65,6 +66,15 @@ export class App extends React.Component {
 		}
 	}
 
+	protected getLastChangedAuthor(): string {
+		return 'Max Mustermann';
```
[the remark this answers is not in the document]
LKuechler: The author should probably be removed completely. There is way to much missing for showing the author. This should just be done with a different story

src/component/page-list/page-tile-composite.tsx
Palmaswell: You are right. The `page-tile-composite.tsx ` is gone :)

src/component/page-list/page-tile-container.tsx
Palmaswell: The `this.inputValue` is a dynamic value that changes with the entered value from the input element. But if the value has not been set, then it should use the current name of the page.

src/component/page-list/page-tile-container.tsx
Palmaswell: When the input element is no longer `:focus` it should just set the `this.editable` flag to false. Changes are not saved.

Is that what you mean @TheReincarnator?

src/component/page-list/page-list-preview.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { Headline } from '../../lsg/patterns/headline';
+import Space, { Size } from '../../lsg/patterns/space';
+
+export interface PageListPreviewProps {
+	headline: string;
+	lastChangedDate: string;
+}
+
+export const PageListPreview: React.StatelessComponent<PageListPreviewProps> = props => (
+	<Space size={[Size.XXXL, Size.XL, Size.XS, Size.XL]}>
```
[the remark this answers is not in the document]
Palmaswell: @LKuechler I actually do think it should be a LSG pattern. The look it's to specific to be a reusable and generic component. Until now only for the page preview.

src/component/page-list/page-tile-composite.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+
+import { observer } from 'mobx-react';
```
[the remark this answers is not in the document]
Palmaswell: The component has been removed. In this case, I guess it does not matter any longer 😃

src/component/page-list/page-tile-composite.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { PreviewTile, PreviewTileProps } from '../../lsg/patterns/preview-tile';
+import Space, { Size } from '../../lsg/patterns/space';
+
+export const PageTileComposite: React.StatelessComponent<PreviewTileProps> = observer(
+	(props): JSX.Element => (
+		<Space size={Size.S}>
```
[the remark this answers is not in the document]
Palmaswell: Spacing is very much part of the composition. LSG patterns should remain as generic as possible. This means without any spaces too

src/component/page-list/page-tile-container.tsx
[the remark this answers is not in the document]
Palmaswell: Done 👍

src/component/page-list/page-tile-container.tsx
Palmaswell: @TheReincarnator you are right and it's gone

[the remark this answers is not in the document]
Palmaswell: @LKuechler you are right and it's gone

src/component/page-list/page-tile-composite.tsx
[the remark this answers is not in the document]
LKuechler: Spacing is styling. If we go by this logic everything can be a component (eg. colors, fonts) and can be done in the composition step.

src/component/page-list/page-tile-container.tsx
TheReincarnator: Yes but at least on Windows, the common patterns is that blurring a rename field commits it, not aborts it. How does the Finder behave?

src/store/project.ts
```diff
@@ -82,6 +82,16 @@ export class Project {
 		return this.id;
 	}
 
+	// TODO
+	public getLastChangedAuthor(): string {
```
Jumace: We don't have any user management at the moment and afaik there is no plan to implement one while we're still on alpha. Based on that I would remove this function as it's now only dead code.

src/component/page-list/page-tile-container.tsx
LKuechler: In the finder the changes are saved on blur

src/store/project.ts
```diff
@@ -82,6 +82,16 @@ export class Project {
...
+		return 'Max Mustermann';
+	}
+
+	// TODO
+	public getLastChangedDate(): Date {
```
Jumace: I don't see that function will make sense at all, at least in the moment.
We will need to prepare more to make that work, which is also not planned at the moment. That's why I would delete that code too, it's dead and will not be used in the near future

src/store/project.ts
```diff
@@ -51,6 +68,8 @@ export class Project {
 	public constructor(properties: ProjectProperties) {
 		this.id = properties.id ? properties.id : Uuid.v4();
 		this.name = properties.name;
+		this.lastChangedAuthor = properties.lastChangedAuthor || 'unknown';
```
Jumace: in line 35 the 'lastChangedAuthor' is set to 'none' if it's not defined, here it's set to 'unknown', we should only do one of both

TheReincarnator: Yes, you are right. I fixed this now.

TheReincarnator: Done.

PR #372, opened 2018-04-12
feat: export to sketch and pdf

TheReincarnator:
This is a pull request to add exporting of the current page to sketch or PDF.
This PR is not complete yet, as it depends on another module not publish to npm yet.

### DIFF
--- a/src/electron/menu.ts
+++ b/src/electron/menu.ts
@@ -8,11 +8,38 @@ import {
 } from 'electron';
 import { ElementLocationCommand } from '../store/command/element-location-command';
 import * as FileExtraUtils from 'fs-extra';
+import { Page } from '../store/page/page';
 import { PageElement } from '../store/page/page-element';
 import * as PathUtils from 'path';
+import { PngExporter } from '../export/png-exporter';
 import * as ProcessUtils from 'process';
 import { Store } from '../store/store';
-const { screen, Menu, shell, app, dialog } = remote;
+const { Menu, shell, app, dialog } = remote;
+
+function getPageFileName(): string {
+	return (Store.getInstance().getCurrentPage() as Page).getName();
+}
+
+function querySaveFilePath(
+	title: string,
+	typeName: string,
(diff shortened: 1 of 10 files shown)


src/electron/menu.ts
```diff
@@ -41,7 +69,7 @@ export function createMenu(): void {
 									designkitPath,
 									PathUtils.join(filePaths[0], 'designkit')
 								);
-								store.openStyleguide(`${filePaths[0]}/designkit`);
+								store.openStyleguide(`${filePaths[0]} / designkit`);
```
Jumace: Are the added spaces here right? As it's a path I'm pretty sure we don't want them here

src/export/sketch-exporter.ts
```diff
@@ -0,0 +1,64 @@
...
+		page.addLayer(pageGroup);
+
+		pageGroup.addLayer(rootLayer);
+
+		FileExtraUtils.writeFileSync(path, JSON.stringify(page.toJSON(), null, '\t'));
+	}
+
+	// tslint:disable-next-line:no-any
+	private static isNotEmpty(value: any): boolean {
+		return value !== undefined && value !== null && value === '';
```
Jumace: You could simplify this code a bit:
`return Boolean(value)`
This works for `undefined`, `''` and `null`

TheReincarnator: Unfortunatelly, Boolean(value) issues `false` for `0`, or is there any good standard "is-empty" check in JavaScript, including strings?

TheReincarnator:
@Jumace No worries, I now split the commit into three, and yes, the spaces in the path were an error, thanks for spotting!

src/export/sketch-exporter.ts
Jumace: Oh okay in that case your code is probably the best solution. Afaik there is no function like `isEmpty`.

marionebl:
🎉

PR #375, opened 2018-04-13
build(scripts): get latest designkit master on build

fahrradflucht:
Get the latest designkit master from github when building a release instead of using the local one.

**Why:**
The latest release (`0.7.0`) includes a more or less broken version of designkit because it has its dependencies messed up which makes it unbuildable today. As far as I can see the version contained in `0.7.0` was never a master of `meetalva/designkit` but just a local version of whom evers machine while building the release.

This is now prevented to happen again 🎉!


LKuechler:
Good idea and also seems to work very good. I was concerned of it taking to long but it's just a small step in the build process.

Thank you :)

Issue #180, opened 2017-12-26
Direct text editing in preview


tilmx:
As a user I want to double-click on a text inside the preview and directly edit the text (instead of editing it in the property pane)

Issue #131, opened 2017-12-21
Direct Preview Manipulation


tilmx:
As a user I should be able to drag and drop a new pattern directly into the preview. The areas where the pattern can be dropped (due to restrictions) should be highlighted.

We have to discuss, how this feature will work in detail.

tilmx:
Splitted into #449 and #450

PR #389, opened 2018-04-20
change to iframe and client/server encapsulation

marionebl:
This PR alters the preview to be a fully independent browser application that can be
exported to HTML more easily and is safer to execute. This also gives us real time demos outside of Alva, e.g. in the Browser or even mobile devices.

* Ensure component library code has no access to Node.js or Electron APIs
* Introduce express.js to create a full local origin to host the preview application on
* Serve JavaScript bundles via webpack and webpack-dev-middleware suitable for browser consumption
* Proxy messages: renderer <->[ipc]<-> main ->[websockets]-> preview

## Drawbacks

* Higher complexity for features that need to pass data into the preview
* App asks for permissions to accept incoming connections if macOS firewall is enabled (if app bundle is unsigned)

## ToDo

* [x] Restore highlight for active elements
* [x] Restore updates for prop updates
* [x] Restore export features
* [x] Restore enum properties
* [x] Find out why styled-components throws sometimes
* [x] Restore synthetic components
* [x] Show errors when rendering fails

## Deferred

I'd like to reimplement the functionality of custom `previewFrame` files in a separate PR.

* [ ] Restore abilities of custom previewFrame

## Commit messages

I'll squash this to `refactor: change to iframe and client/server encapsulation `


tilmx:
Sounds dope! Will this also work with the upcoming drag’n’drop of elements into the preview pane (#131 and #180)?

marionebl:
> Sounds dope! Will this also work with the upcoming drag’n’drop of elements into the preview pane (#131 and #180)?

Sure, there is fast two-way communication between the browser application and the server via web sockets ✨

marionebl:
@LKuechler: Could you outline what the "custom previewFrame" you mention is about?

src/electron/preview.tsx
```diff
@@ -0,0 +1,321 @@
...
+
+interface TreeNode {
+	contents: {
+		[slot: string]: PreviewComponentProps[];
+	};
+	pattern: string;
+}
+
+// tslint:disable-next-line:no-any
+function getComponent(props: any): string | React.SFC<any> | null {
```
LKuechler: It would be great if we could type this. We assume that these props have a pattern and a name property.
But if we look at the place where the function is called the thrown in props are typed as `InjectedPreviewComponentProps` which does not have a pattern or a name property

LKuechler: Maybe just type it as a Partial in cases like these. Still better than any :)

src/electron/preview.tsx
```diff
@@ -0,0 +1,321 @@
...
+	return Object.keys(tree.contents || {})
+		.reduce((acc, key) => [...acc, ...tree.contents[key]], [])
+		.reduce((acc, node) => {
+			deriveComponents(node, acc);
+			return acc;
+		}, init);
+}
+
+// tslint:disable-next-line:no-any
+function scheduleScript(store: any): void {
```
LKuechler: Can this not just be of type `MobX.ObservableMap<{}>`?

src/electron/preview.tsx
```diff
@@ -0,0 +1,321 @@
...
+	const pageId = store.get('pageId');
+
+	// tslint:disable-next-line:no-any
+	const page = (pages as any[]).find(p => p.id === pageId);
+
+	if (page) {
+		const componentRequests = [...deriveComponents(page.root)];
+
+		// tslint:disable-next-line:no-any
+		const components = store.get('components') as any[];
```
LKuechler: `store.get` returns a valid interface it could only be undefined. If you remove the typecast and add a guard for making sure that components is defined we would not need to disable tslint here.

src/electron/preview.tsx
```diff
@@ -0,0 +1,321 @@
...
+		// tslint:disable-next-line:no-any
+		const tasks = store.get('tasks') as any[];
```
LKuechler: Same as above

src/electron/main.ts
```diff
@@ -22,13 +37,69 @@ function createWindow(): void {
...
+						type: 'start-app',
+						payload: port
+					});
+				}
+			}
+		}
+	});
+
+	// tslint:disable-next-line:no-any
+	server.on('client-message', (envelope: any) => {
```
LKuechler: as we assume to get a string here we could also type it as such. Maybe add a comment that a different client message could be triggered and that we need the try/catch because of this

src/export/pdf-exporter.ts
```diff
@@ -1,26 +1,110 @@
...
+
+			webview.addEventListener('did-finish-load', () => {
+				if (webview.src === initial) {
+					return;
 				}
-			);
+				shoot();
+			});
+
+			// tslint:disable-next-line:no-any
```
LKuechler: can be removed

src/export/png-exporter.ts
```diff
@@ -1,55 +1,117 @@
...
+			};
+
+			webview.addEventListener('did-finish-load', () => {
+				if (webview.src === initial) {
+					return;
+				}
+				shoot();
+			});
+
+			// tslint:disable-next-line:no-any
```
LKuechler: can be removed

src/export/png-exporter.ts
```diff
@@ -1,55 +1,117 @@
...
+							height: Math.round(config.height * scaleFactor)
 						},
 						capture => {
 							const pngBuffer: Buffer = capture.toPNG();
 							resolve(pngBuffer);
 
-							// reset the webview height
-							webview.style.height = '100%';
+							setTimeout(() => {
+								// document.body.removeChild(webview);
```
LKuechler: is this one still valid?

src/store/store.ts
```diff
@@ -311,6 +316,41 @@ export class Store {
...
+				const persistedPath = ref.getLastPersistedPath();
+
+				if (!persistedPath) {
+					return null;
+				}
+
+				const pagePath: string = PathUtils.join(pagesPath, persistedPath);
+
+				// tslint:disable-next-line:no-any
+				const data: any = Persister.loadYamlOrJson(pagePath);
```
LKuechler: why is the any needed here?

src/electron/main.ts
```diff
@@ -22,13 +37,69 @@ function createWindow(): void {
...
-		url.format({
-			pathname: PathUtils.join(__dirname, 'renderer', 'app.html'),
+		Url.format({
+			pathname: Path.join(__dirname, 'app.html'),
 			protocol: 'file:',
 			slashes: true
 		})
 	);
 
+	// tslint:disable-next-line:await-promise
```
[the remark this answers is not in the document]
marionebl: Addressed via casting `PromiseLike<number>` to `Promise<number>`

src/electron/preview.tsx
marionebl: Adressed by introducing a strongly typed `PreviewStore`

src/electron/preview.tsx
marionebl: Adressed by introducing a strongly typed `PreviewStore`

src/electron/preview.tsx
marionebl: Adressed by introducing a strongly typed `PreviewStore`

src/electron/server.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export async function createServer(opts: ServerOptions): Promise<EventEmitter> {
+	const store = Store.getInstance();
+	store.openFromPreferences();
+
+	const emitter = new EventEmitter();
+	const app = express();
+
+	// tslint:disable-next-line:no-any
+	const server = Http.createServer(app as any);
```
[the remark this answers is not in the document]
marionebl: `components` in fact is `string[]`. I added a `PreviewStore`, so this should get clearer

src/export/pdf-exporter.ts
marionebl: done

marionebl:
@LKuechler Adressed all concerns!

src/electron/preview.tsx
```diff
@@ -0,0 +1,330 @@
...
+function main(): void {
+	SmoothscrollPolyfill.polyfill();
+
+	const store = MobX.observable.map({});
+	store.set('components', []);
+	store.set('tasks', []);
+
+	const highlight = new HighlightArea();
+
+	const connection = new WebSocket(`ws://${window.location.host}`);
```
fadiguezel: Did you think about using wss which would encrypt all the traffic between the server and the iFrame?

src/electron/server.ts
```diff
@@ -0,0 +1,139 @@
...
+					// tslint:disable-next-line:no-any
+				} as any);
+
+				app.use('/scripts', webpackDevMiddleware(compiler));
+			}
+		}
+
+		wss.clients.forEach(client => {
+			if (client.readyState === OPEN) {
+				client.send(JSON.stringify(message));
```
fadiguezel: all clients share the same session right?
also a rouge web socket connecting to the server?

marionebl: Yes, that's the case. Any suggestions to improve on this?

src/electron/preview.tsx
marionebl: Yes, but currently we do not have a good way to produce certificates on the fly. Any suggestions?

marionebl:
> Nice, really cool! I just stumbled across that when changing enum properties the preview doesn’t react to that. String and boolean properties worked well.

enum props worked in my latest tests, could you check again.

src/electron/main.ts
```diff
@@ -22,13 +37,69 @@ function createWindow(): void {
 
 	// and load the index.html of the app.
 	win.loadURL(
-		url.format({
-			pathname: PathUtils.join(__dirname, 'renderer', 'app.html'),
+		Url.format({
+			pathname: Path.join(__dirname, 'app.html'),
```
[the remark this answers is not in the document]
marionebl: Ok, let's do this in a separate PR as discussed

src/export/png-exporter.ts
```diff
@@ -1,55 +1,116 @@
...
+				webview.loadURL(
+					`data:text/html;charset=utf-8,${encodeURIComponent(payload.document)}`,
+					{
+						baseURLForDataURL: payload.location
+					}
+				);
+			};
+
+			// (3) Wait for webview to be ready and capture the page
+			const shoot = () => {
```
TheReincarnator: Same as above.

src/export/sketch-exporter.ts
```diff
@@ -1,31 +1,30 @@
-import * as HtmlSketchApp from '@brainly/html-sketchapp';
-import { ipcRenderer, WebviewTag } from 'electron';
+import { ipcRenderer } from 'electron';
 import { Exporter, ExportResult } from './exporter';
 import { Page } from '../store/page/page';
 import { Store } from '../store/store';
 import * as uuid from 'uuid';
 
-type Message = SuccessMessage | ErrorMessage;
+// type Message = SuccessMessage | ErrorMessage;
```
TheReincarnator: Please delete commented-out code.

src/store/page/page-ref.ts
```diff
@@ -23,7 +23,7 @@ export class PageRef {
 	/**
 	 * The technical (internal) ID of the page.
 	 */
-	@MobX.observable private id: string;
+	@MobX.observable public id: string;
```
TheReincarnator: Why this? This is not good and not necessary IMHO.

src/store/page/page.ts
```diff
@@ -135,7 +135,10 @@ export class Page {
 	 * @return The JSON object to be persisted.
 	 */
 	public toJsonObject(): JsonObject {
-		return { root: this.root ? this.root.toJsonObject() : undefined };
+		return {
```
TheReincarnator: This changes the persisted files. Why is it necessary?

src/electron/main.ts
```diff
@@ -22,13 +37,69 @@ function createWindow(): void {
...
+
+	// tslint:disable-next-line:no-any
+	server.on('client-message', (envelope: any) => {
+		try {
+			const message = JSON.parse(envelope);
+			switch (message.type) {
+				case 'content-response':
+					send({
+						id: message.id,
+						payload: message.payload,
```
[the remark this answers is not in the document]
marionebl: Yes, it does. Let's do this in a separate PR and for both directions

src/styleguide/renderer/react/preview.tsx
```diff
@@ -1,321 +0,0 @@
...
-	protected createComponent(pageElement: PageElement): React.ReactNode {
-		const pattern = pageElement.getPattern();
-		if (!pattern) {
-			return null;
-		}
-
-		try {
-			const patternId: string = pattern.getId();
-			if (patternId === Pattern.SYNTHETIC_TEXT_ID) {
-				return this.createStringComponent(pageElement);
```
TheReincarnator: The string and asset synthetic components got lost on the way.
The Placeholder was a separate tsx file, while string was just... as string.
Both feature no longer work in the new solution.

src/electron/main.ts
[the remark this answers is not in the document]
TheReincarnator: Now done in this PR.

src/electron/preview.tsx
fadiguezel: Good point. We should rethink the effort vs security threat gain.

src/styleguide/renderer/react/preview.tsx
```diff
@@ -1,321 +0,0 @@
...
-				const slotContents = pageElement.getSlotContents(slotId);
-				componentProps[reactPropertyName] = slotContents.map((child, index) =>
-					this.createComponent(child)
-				);
-			});
-		}
-	}
-
-	// tslint:disable-next-line:no-any
-	protected collectPropertyValues(componentProps: any, pageElement: PageElement): void {
```
TheReincarnator: This method is missing in the new solution and the reason why enums fail.
All property values must be converted using property.convertToRender.

marionebl: Resolved via 6170e7615862569194e3b8e656a52a265ef90e30

marionebl: Appears to be still broken, could you have another look @TheReincarnator?

marionebl: I tested with the first `<Headline/>` - changing `Level` and `TextAlign` does nothing. Changing the `Size` of a `<Space>`, does though.

src/electron/main.ts
```diff
@@ -22,13 +38,69 @@ function createWindow(): void {
...
 		})
 	);
 
+	// Cast getPort return type from PromiseLike<number> to Promise<number>
+	// to avoid async-promise tslint rule to produce errors here
+	const port = await (getPort({ port: 1879 }) as Promise<number>);
+	const server = await createServer({ port });
+
+	// tslint:disable-next-line:no-any
+	const send = (message: any) => {
```
LKuechler: We could type this a bit:
`const send = (message: { id: string; payload: any; type: string }) => {...`
Maybe also put these typings in there own interface

src/electron/preview.tsx
```diff
@@ -0,0 +1,278 @@
...
+				setTimeout(() => {
+					props.store.elementId = '';
+				}, 500);
+			}
+		}
+	}
+
+	public render(): JSX.Element | null {
+		const props = this.props as InjectedPreviewComponentProps;
+		const contents = props.contents ? props.contents : {};
```
LKuechler: could also just be `props.contents || {}`

src/electron/renderer.ts
```diff
@@ -0,0 +1,80 @@
+import { App } from '../component/container/app';
+import { ipcRenderer, webFrame } from 'electron';
+import * as MobX from 'mobx';
+// import { Page } from '../store/page/page';
```
LKuechler: can be removed

src/export/sketch-exporter.ts
```diff
@@ -1,31 +1,30 @@
...
 
 interface ErrorMessage {
 	error: Error;
 	id: string;
-}
+} */
 
 export class SketchExporter extends Exporter {
-	private id: string;
+	// private id: string;
```
LKuechler: Should this be removed?

src/export/sketch-exporter.ts
```diff
@@ -1,31 +1,30 @@
...
 import { Exporter, ExportResult } from './exporter';
 import { Page } from '../store/page/page';
 import { Store } from '../store/store';
 import * as uuid from 'uuid';
 
-type Message = SuccessMessage | ErrorMessage;
+// type Message = SuccessMessage | ErrorMessage;
 
-interface SuccessMessage {
+/* interface SuccessMessage {
```
LKuechler: Should this be removed?

src/export/sketch-exporter.ts
```diff
@@ -1,31 +1,30 @@
...
+} */
 
 export class SketchExporter extends Exporter {
-	private id: string;
+	// private id: string;
 
 	public constructor() {
 		super();
-		this.id = uuid.v4();
+		// this.id = uuid.v4();
```
LKuechler: same as above

src/export/sketch-exporter.ts
```diff
@@ -1,31 +1,30 @@
...
 	}
 
-	public static createSketchExport(): string {
+	/* public static createSketchExport(): string {
```
LKuechler: same as above

LKuechler:
All the exports seem to be broken at the moment. Could some one validate that this is not just a problem on my machine?

marionebl:
> All the exports seem to be broken at the moment. Could some one validate that this is not just a problem on my machine?

Fixed via afd70f05f5cf5c21ba8427f83c48213c23ed8510

src/styleguide/renderer/react/preview.tsx
marionebl: Scratch that, this is fixed.

tilmx:
I’m not sure why, but you can’t drag elements between other elements any more... Dropping it as a child works, but the in-between drop doesn’t work. :-(

src/electron/server.ts
```diff
@@ -0,0 +1,297 @@
...
+	payload: {
+		elementId?: string;
+		// tslint:disable-next-line:no-any
+		page?: any;
+	};
+	type: 'state';
+}
+
+// tslint:disable-next-line:no-any
+type Queue = any[];
```
TheReincarnator: Why `any`? Can't we define an interface? We don't like `any`s... :-)

src/electron/server.ts
```diff
@@ -0,0 +1,297 @@
...
+
+	const styleguide = new Styleguide(
+		update.styleguidePath,
+		update.patternsPath,
+		update.analyzerName
+	);
+
+	const context = styleguide.getPath();
+
+	const components = styleguide.getPatterns().reduce((acc, pattern) => {
```
TheReincarnator: What is acc?

src/lsg/patterns/panes/preview-pane/index.tsx
```diff
@@ -4,13 +4,14 @@ import * as React from 'react';
 import styled from 'styled-components';
 
 export interface PreviewPaneProps {
-	handleMouseDownLeft?: React.MouseEventHandler<HTMLElement>;
-	handleMouseDownRight?: React.MouseEventHandler<HTMLElement>;
-	handleMouseMove?: React.MouseEventHandler<HTMLElement>;
-	handleMouseUp?: React.MouseEventHandler<HTMLElement>;
+	id?: string;
+	onMouseDownLeft?: React.MouseEventHandler<HTMLElement>;
```
TheReincarnator: We had a discussion on when to use "handle", and when to use "on", but I forgot the outcome. @LKuechler ?

src/electron/loader.ts
```diff
@@ -0,0 +1,27 @@
+import { getOptions } from 'loader-utils';
+import { loader } from 'webpack';
+
+// No typings available for commondir
+const commondir = require('commondir');
+
+module.exports = alvaEntryLoader;
```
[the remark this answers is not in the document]
marionebl: TypeScript transpiles `export default thing` to `module.exports.default = thing`. This `loader` is required by `webpack`, which expects the loader function to be exported on `module.exports` directly.

src/electron/safe-pattern.ts
```diff
@@ -0,0 +1,8 @@
+export function safePattern(id: string): string {
```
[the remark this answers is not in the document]
marionebl: This lives in a separate file so we can use it both on the server and preview. I suggest `patternIdToSlug` or something similiar, your call though - no stakes in any specific name for this.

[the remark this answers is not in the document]
TheReincarnator: That's better, but I'd like to replace slug by something meaningful. patternIdToWebpackName?

src/lsg/patterns/panes/preview-pane/index.tsx
marionebl: To my knowledge the React community consensus is `handle*` for internal methods and `on*` for property names.

PR #393, opened 2018-04-23
feat: update styling of layout and element pane

tilmx:
Bringing the chrome, preview and element pane to the freshest style version!

The pull request will be squashed into "feat: update styling of layout and element pane"


src/lsg/patterns/element/index.tsx
```diff
@@ -45,69 +45,116 @@ export interface StyledPlaceholder {
...
 			background: ${colors.grey90.toString()};
+
+			&::before {
+				background: ${colors.grey90.toString()};
+			}
 		`
 			: ''};
 `;
 
 const StyledPlaceholder = styled.div`
```
tilmx: @LKuechler Is it possible to show the Placeholder ONLY when an element is being dragged around?

src/lsg/patterns/element/index.tsx
```diff
@@ -45,69 +45,116 @@ export interface StyledPlaceholder {
 }
 
 const StyledElement = styled.div`
-	cursor: default;
 	position: relative;
+	z-index: 1;
 `;
 
 const StyledElementLabel = styled.div`
```
tilmx: @LKuechler Is it possible to apply styling on the element during dragging around? So like opacity or scale?

tilmx: Solved with 71c298d81a3a4fa0c4f046c6a159f704028ce5b9

src/lsg/patterns/element/index.tsx
tilmx: Solved with 4f6f148

src/component/container/app.tsx
```diff
@@ -145,54 +140,51 @@ export class App extends React.Component {
...
+						<React.Fragment>
+							<SideBar
+								key="left"
+								directionVertical
+								onClick={() => store.setSelectedElement()}
+							>
+								<ElementPane>
+									<ElementList />
+									<AddButton
+										active={store.getRightPane() === RightPane.Patterns}
```
marionebl: Shouldn't this be invisible / disabled if the right sidebar displays `RightPane.Patterns`?

src/lsg/patterns/add-button/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+interface StyledAddButtonProps {
+	active?: boolean;
+}
+
+interface StyledIconProps {
+	active?: boolean;
+}
+
+const StyledAddButton = styled.div`
+	position: absolute;
```
marionebl: Should be `position: fixed`, currently this happens: 
screen shot 2018-04-25 at 18 42 08

src/lsg/patterns/chrome/index.tsx
```diff
@@ -20,12 +20,15 @@ const StyledChrome = styled.div`
 	display: flex;
 	align-items: center;
 	width: 100%;
-	height: 54px;
+	height: 40px;
 	padding: ${getSpace(SpaceSize.XS)}px ${getSpace(SpaceSize.XXL) * 3}px;
+	border-bottom: 0.5px solid ${colors.grey90.toString()};
```
marionebl: We should not drop on subpixels - this might produce undesirable output on devices with lower pixel density.
This should be guarded with a  media query, like this:

```
border-bottom-width: 1px;
@media screen and (-webkit-min-device-pixel-ratio: 2) {
  border-bottom-width: 0.5px;
}
```

src/lsg/patterns/element/index.tsx
```diff
@@ -122,78 +173,61 @@ const StyledElementChild = styled.div`
...
-	width: 12px;
-	height: 12px;
+	width: ${getSpace(Size.S)}px;
+	height: ${getSpace(Size.S)}px;
 	padding: ${getSpace(Size.XS)}px;
 	transition: transform 0.2s;
 
 	${(props: StyledIconProps) => (props.open ? 'transform: rotate(90deg)' : '')};
-	${(props: StyledIconProps) => (props.active ? 'fill: white' : '')};
+	${(props: StyledIconProps) => (props.active ? 'fill: #0070D6' : '')};
```
marionebl: This should be deduced from styleguide colors

src/lsg/patterns/list/index.tsx
```diff
@@ -44,9 +44,8 @@ const StyledUl = styled.ul`
 const StyledLi = styled.li`
 	line-height: 25px;
 	list-style: none;
-	${(props: StyledListItemProps) => (props.onClick ? 'cursor: pointer;' : '')} ${(
-		props: StyledListItemProps
-	) => (props.active ? 'background: #def' : '')};
+	${(props: StyledListItemProps) => (props.onClick ? 'cursor: pointer;' : '')};
+	${(props: StyledListItemProps) => (props.active ? 'background: #def' : '')};
```
marionebl: Move the colors pattern

src/component/container/element-list.tsx
```diff
@@ -208,40 +226,134 @@ export class ElementList extends React.Component {
...
+		}
+
+		const target = e.target as HTMLElement;
+		const dragElement = target.querySelector('div');
+
+		// restyle the drag image and move it somewhere invisible
+		if (dragElement) {
+			const dragImg = dragElement.cloneNode(true) as HTMLElement;
+			dragImg.setAttribute('style', DRAG_IMG_STYLE);
+			document.body.appendChild(dragImg);
```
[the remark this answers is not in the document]
tilmx: I get the idea, but what do mean with "this.dragImg = dragImg;" instead of "e.dataTransfer.setDragImage(dragImg, 75, 15);"? Now it doesn’t set the drag image anymore. Or did you mean to use it in a variable then?

src/electron/main.ts
```diff
@@ -17,6 +17,7 @@ function createWindow(): void {
 		minWidth: 780,
 		minHeight: 380,
 		titleBarStyle: 'hiddenInset',
+		backgroundColor: '#f7f7f7',
```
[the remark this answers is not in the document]
tilmx: Good catch. In "global-styles" that makes sense – but as an open question: Wouldn’t it be better to directly set the color here in main.ts in terms of startup speed?

src/lsg/patterns/add-button/index.tsx
tilmx: Good catch, thanks!

src/component/container/element-list.tsx
```diff
@@ -10,9 +10,30 @@ import { Pattern } from '../../store/styleguide/pattern';
...
+const DRAG_IMG_STYLE = `
+	position: absolute;
+	background-color: #fff;
+	color: #000;
+	padding: 6px 18px;
+	border-radius: 3px;
+	font-size: 12px;
+	opacity: 1;
+	top: 0;
+	left: -500px;
```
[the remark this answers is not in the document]
tilmx: Good point. Thanks!

src/electron/main.ts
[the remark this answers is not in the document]
marionebl: Sure, nothing wrong with applying a background color here, but I'd recommend we the actual value from the colors pattern.

src/component/container/element-list.tsx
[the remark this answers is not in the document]
marionebl: Sorry, I emitted the `e.dataTransfer.setDragImage(dragImg, 75, 15);` statement, that's still needed. `this.dragImg` should happen right after `e.dataTransfer.setDragImage(dragImg, 75, 15);`;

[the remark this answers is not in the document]
tilmx: Solved with 70d4708

src/component/container/app.tsx
```diff
@@ -145,54 +140,51 @@ export class App extends React.Component {
...
+										label="Add Elements"
+										onClick={e => {
+											e.stopPropagation();
+											store.setRightPane(RightPane.Patterns);
+											store.setSelectedElement();
```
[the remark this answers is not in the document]
tilmx: As interacting with the left layers changes the right property pane, the idea was to add a button that shows the pattern view. And the short cut for not having to directly click on the button left bottom would be to click on the unused area in the element view (so it unselects the element). Then you always see on the right side what you want to do (like inspect on macOS does).

I’d love to test this in a kind of usertest, if we find out it doesn’t work – let’s change it.

src/component/container/app.tsx
```diff
@@ -145,54 +140,51 @@ export class App extends React.Component {
...
+					{project ? (
+						<React.Fragment>
+							<SideBar
+								key="left"
+								directionVertical
+								onClick={() => store.setSelectedElement()}
+							>
+								<ElementPane>
+									<ElementList />
+									<AddButton
```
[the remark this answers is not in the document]
tilmx: I thought an active state would be cooler, to not disturb the user because the button is sometimes not there. But let’s test that as well (see below)!

PR #395, opened 2018-04-23
fix(object-property): infinite loop on type recursion

faselbaum:
Object types have been expanded at creation time which means that the following code within a components prop type declarations would lead to an ifinite loop.

```
export interface FooProps {
  test: string;
  bar: BarProps;
}

export interface BarProps {
  foo: FooProps;
}
```

Object properties will now be expanded at render time (only if the user expands them in the property pane).

### DIFF
--- a/src/store/styleguide/property/object-property.ts
+++ b/src/store/styleguide/property/object-property.ts
@@ -1,5 +1,7 @@
 import { Property } from './property';
 
+export type PropertyResolver = () => Property[];
+
 /**
  * An object property is a property that supports objects with nested property values.
  * The properties field declares the types of each object element.
@@ -11,7 +13,9 @@ export class ObjectProperty extends Property {
 	/**
 	 * The nested properties this property supports in its object values.
 	 */
-	private properties: Map<string, Property> = new Map();
+	private properties?: Map<string, Property>;
+
+	private propertyResolver?: PropertyResolver;
 
 	/**
 	 * Creates a new object property.
@@ -39,7 +43,7 @@ export class ObjectProperty extends Property {
 	 * @return The nested properties this property supports.
 	 */
 	public getProperties(): Property[] {
-		return Array.from(this.properties.values());
+		return Array.from(this.resolveProperties().values());
 	}
 
 	/**
@@ -48,7 +52,7 @@ export class ObjectProperty extends Property {
 	 * @return The nested property if the ID was found.
 	 */
 	public getProperty(id: string): Property | undefined {
-		return this.properties.get(id);
+		return this.resolveProperties().get(id);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


src/styleguide/analyzer/typescript-react-analyzer/property-analyzer.ts
```diff
@@ -214,7 +214,10 @@ export class PropertyAnalyzer {
 
 			if (objectType.objectFlags & ts.ObjectFlags.Interface) {
 				const property = new ObjectProperty(args.symbol.name);
-				property.setProperties(PropertyAnalyzer.analyze(args.type, args.typechecker));
+				// tslint:disable-next-line:arrow-return-shorthand
+				property.setPropertyResolver(() => {
```
LKuechler: instead of disabling the linting rule you could just write this:
`property.setPropertyResolver(() => PropertyAnalyzer.analyze(args.type, args.typechecker))`

faselbaum: oh yes sorry i forgot that line. i had a debug breakpoint there before :)
thx for finding it

src/store/styleguide/property/object-property.ts
```diff
@@ -58,16 +62,29 @@ export class ObjectProperty extends Property {
...
+		if (!this._properties) {
+			if (!this.propertyResolver) {
+				throw new Error('property resolver is not set');
+			}
+
+			const properties = this.propertyResolver();
+			this._properties = new Map();
+
+			// tslint:disable-next-line:no-non-null-assertion
+			properties.forEach(property => this._properties!.set(property.getId(), property));
```
LKuechler: It is a bit strange that you have to use a `!` here.
Could we not just make the `_properties` non optional and have it always be a empty Map.
Then you would not have to disable tslint for this line :)

src/store/styleguide/property/object-property.ts
```diff
@@ -11,7 +13,9 @@ export class ObjectProperty extends Property {
 	/**
 	 * The nested properties this property supports in its object values.
 	 */
-	private properties: Map<string, Property> = new Map();
+	private _properties?: Map<string, Property>;
```
TheReincarnator: We never used underscore prefixes before and I'd like to vote against them.
Properties should always be private, and their access by getters/setters, so we don't need to distinguish any privacy here.

src/store/styleguide/property/object-property.ts
TheReincarnator: I guess @faselbaum is trying to use "undefined" as a lazy indicator.
Alternative would be to use another const and assign it at the end of the method. This would also ensure that the data structure is not half-constructed if there is a problem.

TheReincarnator:
Review done, only minor comments.

TheReincarnator:
My comments are ok now, thanks!

PR #399, opened 2018-04-24
(feat): Page Switch

Palmaswell:
Page Switch implementation


src/component/chrome/chrome-container.tsx
```diff
@@ -24,13 +27,17 @@ export class ChromeContainer extends React.Component {
...
-				onLeftClick={() => (previousPage ? store.openPage(previousPage.getId()) : undefined)}
-				onRightClick={() => (nextPage ? store.openPage(nextPage.getId()) : undefined)}
-				title={page ? page.getName() : undefined}
-			>
+			<Chrome>
+				<OverviewSwitchContainer />
+				<ViewSwitch
+					fontSize={Size.M}
+					justify="center"
+					leftVisible={!!previousPage}
```
LKuechler: using `Boolean(previousPage)` would be a bit nicer

src/lsg/patterns/view-switch/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Icon, IconName, IconProps, Size as IconSize } from '../icons';
+import * as React from 'react';
+import { getSpace, Size as SpaceSize } from '../space';
+import styled from 'styled-components';
+
+export interface ViewSwitchProps {
+	fontSize?: Size;
+	justify?: 'start' | 'center' | 'end' | 'stretch';
+	leftVisible: boolean;
+	onLeftClick: React.MouseEventHandler<SVGElement>;
```
LKuechler: please name all event handler `handle...`. This way they are a lot easier to differentiate from the native react event handlers

src/component/chrome/chrome-container.tsx
```diff
@@ -24,13 +27,17 @@ export class ChromeContainer extends React.Component {
...
+					rightVisible={!!nextPage}
+					onLeftClick={() => (previousPage ? store.openPage(previousPage.getId()) : undefined)}
```
LKuechler: `onLeftClick` is not optional. It should not be valid to pass undefined as a value here.
Maybe just always pass the function. I will not be triggered because the buttons are not visible if not needed

src/lsg/patterns/view-switch/index.tsx
```diff
@@ -0,0 +1,80 @@
+import { colors } from '../colors';
+import { Size } from '../copy';
+import { Icon, IconName, IconProps, Size as IconSize } from '../icons';
+import * as React from 'react';
+import { getSpace, Size as SpaceSize } from '../space';
+import styled from 'styled-components';
+
+export interface ViewSwitchProps {
+	fontSize?: Size;
+	justify?: 'start' | 'center' | 'end' | 'stretch';
```
LKuechler: This is not very nice. You are just moving every css possibility in to the props.

src/component/chrome/chrome-container.tsx
```diff
@@ -24,13 +27,17 @@ export class ChromeContainer extends React.Component {
...
-				rightVisible={!!nextPage}
-				onLeftClick={() => (previousPage ? store.openPage(previousPage.getId()) : undefined)}
-				onRightClick={() => (nextPage ? store.openPage(nextPage.getId()) : undefined)}
-				title={page ? page.getName() : undefined}
-			>
+			<Chrome>
+				<OverviewSwitchContainer />
+				<ViewSwitch
+					fontSize={Size.M}
+					justify="center"
```
Jumace: please create a type for `justify` in `src/lsg/patterns/view-switch/index.tsx` and export it so it can be used here

src/component/chrome/chrome-container.tsx
Jumace: please refactror this too to use `!!` but `Boolean()`. The `!!` is pretty easy to oversee but `Boolean()` is easy to see. Afaik that is also something that was agreed on for this project. Correct me if I'm wrong

src/component/chrome/chrome-container.tsx
```diff
@@ -24,13 +27,17 @@ export class ChromeContainer extends React.Component {
...
+					leftVisible={!!previousPage}
+					rightVisible={!!nextPage}
```
Jumace: please refactror this too to use `!!` but `Boolean()`. The `!!` is pretty easy to oversee but `Boolean()` is easy to see. Afaik that is also something that was agreed on for this project. Correct me if I'm wrong

src/component/chrome/overview-switch-container.tsx
```diff
@@ -0,0 +1,32 @@
+// import * as MobX from 'mobx';
```
Jumace: if you don't need this one please delete the line

src/component/chrome/overview-switch-container.tsx
```diff
@@ -0,0 +1,32 @@
...
+			return project ? project.getName() : 'Unnamed Project';
+		}
+	}
+
+	public render(): JSX.Element {
+		const store = Store.getInstance();
+		return (
+			<ViewSwitch
+				onLeftClick={() => store.togglePageOverview()}
+				onRightClick={() => null}
```
Jumace: if you don't need that function always, please consider to make it optional

src/lsg/patterns/chrome/demo.tsx
```diff
@@ -0,0 +1,30 @@
+import { Size } from '../copy';
```
Jumace: `import { Size } from '../../lsg/patterns/copy';` => `import { Size as FontSize } from '../../lsg/patterns/copy';`
With that it is more clear which size is meant. In the copy component it's kind of obvious but not here in the chrome-container

src/lsg/patterns/view-switch/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+	visible: boolean;
+}
+
+interface StyledViewSwitchProps {
+	fontSize?: Size;
+	justify?: 'start' | 'center' | 'end' | 'stretch';
+}
+
+const StyledViewSwitch = styled.div`
+	display: inline-flex;
```
Jumace: doesn't have to be `inline-flex`, `flex` only is enough

src/lsg/patterns/icons/index.tsx
```diff
@@ -41,6 +42,7 @@ const icons: { readonly [key: string]: JSX.Element[][] | JSX.Element[] } = {
 		[<path key="arrow" d="M17.5 12l-8.486 8.485L7.6 19.071 14.671 12 7.6 4.929l1.414-1.414z" />]
 	],
 	[IconName.ArrowFill]: [[<path key="arrowFill" d="M8 4l8 8-8 8z" />]],
```
Jumace: please refactor this one the `ArrowFillRight`with that it will be easier to use the correct one

src/lsg/patterns/view-switch/index.tsx
```diff
@@ -0,0 +1,80 @@
+import { colors } from '../colors';
+import { Size } from '../copy';
```
Jumace: `import { Size } from '../../lsg/patterns/copy';` => `import { Size as FontSize } from '../../lsg/patterns/copy';`
With that it is more clear which size is meant. In the copy component it's kind of obvious but not here in the chrome-container

Jumace:
When I tried the feature the first thing I tried was to click on `Meet Alva Overview` so the "label" but nothing happened. In my opinion the label should be clickable too. @markusoelhafen maybe that is more a question for you.

src/component/chrome/chrome-container.tsx
Palmaswell: Done ✅

src/lsg/patterns/view-switch/index.tsx
Palmaswell: What I've so far understood is that the convention is:
1. onClick: for elements just receiving or passing the event, like in this case the `view-switch`
2. handleClick: for the component that is actually operating or acting on the click event.

What is your opinion about that?

LKuechler: Currently we only use `on` when we use a pass something to a native react event prop.
This has the benefit that you can immediately see which event handlers are custom ones and could have a different response than you would assume.

src/component/chrome/chrome-container.tsx
Palmaswell: Done ✅

src/component/chrome/chrome-container.tsx
```diff
@@ -1,8 +1,11 @@
 import Chrome from '../../lsg/patterns/chrome';
+import { Size } from '../../lsg/patterns/copy';
```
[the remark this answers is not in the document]
Palmaswell: Done ✅

src/component/chrome/chrome-container.tsx
Palmaswell: Done ✅

src/lsg/patterns/view-switch/index.tsx
Palmaswell: Well, that is right. When more items come in chrome we will be going to need them :).
Would you like me to reduce them to the current necessary?

src/component/chrome/chrome-container.tsx
Palmaswell: Done ✅

src/component/chrome/chrome-container.tsx
Palmaswell: Done ✅

src/component/chrome/overview-switch-container.tsx
Palmaswell: sorry, my bad. Done too :)

src/component/chrome/overview-switch-container.tsx
Palmaswell: ✅

src/lsg/patterns/chrome/demo.tsx
Palmaswell: ✅

src/lsg/patterns/view-switch/index.tsx
Palmaswell: actually in this case you're right. ✅

src/lsg/patterns/view-switch/index.tsx
Palmaswell: ✅

src/lsg/patterns/icons/index.tsx
Palmaswell: ✅

Palmaswell:
@Jumace , @markusoelhafen , @tilmx 

Regarding @Jumace comment regarding the clickable label. I think he is right, but if we decide to do so please make a new ticket for this feature. This is a behavior that is already in `Master`

Palmaswell:
@tilmx Regarding your comment. 
1. We decided that the page item should behave the same way files behave in Finder. In Finder when you double click on the file name, you actually open the document.

2. Please reply to @LKuechler about his suggestion, which I think its a very good idea. But we need your feedback in order to implement the functionality.

src/lsg/patterns/view-switch/index.tsx
LKuechler: I think for now we should reduce them :)

LKuechler:
@tilmx **Page Switch in Chrome:**
There could be something that changes when switching pages on the page-overview page. The selected page on the page-overview could change when the currently active page is changed.

tilmx:
Hey @Palmaswell! What’s the status here? Can we finish this PR this week?
Thanks, Tilman

PR #403, opened 2018-04-25
fix(debug): vscode debug profiles on windows

faselbaum:


LKuechler:
commit type should not be `fix`. `chore` would be a better type :)

PR #409, opened 2018-05-02
fix(preview-renderer): rendering of slot contents

faselbaum:


src/electron/preview-renderer.tsx
```diff
@@ -120,8 +119,16 @@ export function render(init: RenderInit): void {
 		public render(): JSX.Element | null {
 			const props = this.props as InjectedPreviewComponentProps;
 			const contents = props.contents || {};
-			const children = typeof contents.default === 'undefined' ? [] : contents.default;
-			const slots = omit(contents, ['default']);
+
+			const renderedContents: { [slotId: string]: React.ReactNode[] } = {};
+			Object.keys(contents).forEach(slotId => {
```
marionebl: Please use `reduce` for simplicity here, also make it clear that `default` lands in `children`

```js
const renderedSlots = Object.keys(omit(contents, ['default'])).reduce((renderedSlots, slotName) => {  
  //  and so forth
});
```

src/electron/preview-renderer.tsx
```diff
@@ -147,9 +154,11 @@ export function render(init: RenderInit): void {
 
 			return (
 				<ErrorBoundary name={props.name}>
-					<Component {...slots} {...props.properties} data-sketch-name={props.name}>
-						{children.map(child => <PreviewComponent key={child.uuid} {...child} />)}
-					</Component>
+					<Component
+						{...props.properties}
+						{...renderedContents}
+						data-sketch-name={props.name}
```
marionebl: Please leave the distinct `children` / `slots.default` handling in place to signal the slot lands in `children`

PR #410, opened 2018-05-02
feat: update styling of preview resizing

markusoelhafen:
clean up the preview pane ui and make resize elements visible on hover.
let me know what you think @tilmx


tilmx:
I really like the styling and the UX itself. The only thing I am a bit unsure is about if it is so good to have those spaces not clickable for the preview content itself. So e.g. if on the website there is something to be clicked on the left or right side, then you can’t click that any more because of that handles. But let’s talk later about that @markusoelhafen!

PR #413, opened 2018-05-02
feat(store): prepare basic events and variables

TheReincarnator:
This PR adds a new property of type 'event', supporting the first two basic event actions:
- Set a variable from an input, and
- Go to a configured page.
The design kit already contains an example, and the two actions work, but the properties panel is lacking the UI for this new property type, so you can't yet change it in the UI.


TheReincarnator:
Revoking PR, will reissue new PR after rebasing against HTML export.

PR #417, opened 2018-05-03
feat: rename page elements #205

### LINKED ISSUE #205: Name Components in Element Pane
It's pretty confusing to have multiple same components (e.g. Layout) underneath each other in the element pane.

It would be good to have the possibility to name the components (e.g. Layout - "Header Section", "Content Section", ...)

marionebl:
This implements an edit mode for page elements in the tree view. 

## Start edit mode

* Click on the label of a selected item
* Hit enter when an item is selected

## Exit edit mode 

* Blur the element
* Select another element
* Hit enter

<details>
<summary>Screencast</summary>

name-element
</details>

Fixes #205

### DIFF
--- a/src/component/container/element-list.tsx
+++ b/src/component/container/element-list.tsx
@@ -169,8 +169,6 @@ export class ElementList extends React.Component<{}, ElementListState> {
 		const store = Store.getInstance();
 		const label = above(e.target, `[${ElementAnchors.label}]`);
 
-		console.log(element, label);
-
 		if (!element) {
 			return;
 		}
--- a/src/component/container/element-list.tsx
+++ b/src/component/container/element-list.tsx
@@ -1,9 +1,9 @@
 import { colors } from '../../lsg/patterns/colors';
 import { elementMenu } from '../../electron/context-menus';
-import { ElementAnchors } from '../../lsg/patterns/element';
+import { ElementAnchors, ElementProps } from '../../lsg/patterns/element';
 import { ElementLocationCommand } from '../../store/command/element-location-command';
+import { ElementNameCommand } from '../../store/command/element-name-command';
 import { ElementWrapper } from './element-wrapper';
-import { ListItemProps } from '../../lsg/patterns/list';
 import { createMenu } from '../../electron/menu';
 import { observer } from 'mobx-react';
 import { Page } from '../../store/page/page';
@@ -33,6 +33,7 @@ const DRAG_IMG_STYLE = `
 @observer
 export class ElementList extends React.Component<{}, ElementListState> {
 	private dragImg?: HTMLElement;
+	private globalKeyUpListener?: (e: KeyboardEvent) => void;
 
 	public state = {
 		dragging: true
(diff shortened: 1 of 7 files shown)


marionebl:
>  Could we blur the input when I hit [Enter]?

Do you mean disabling edit mode **and** unselecting the element?

marionebl:
Adressed all concerns, let me know what you thin @tilmx

tilmx:
Just stumbled across one bug: When I drag an element around and drop it on itself – it deletes itself. We already had this bug somewhen in the beginning and it looks like it just came back. If we get that fixed – I’m really lucky! Or is this a topic for a separate PR?

tilmx:
> And a last thingie: Can we use text-overflow: ellipsis; on page element names that are too long to display? Otherwise the element hight jumps when switching between edit mode and display mode.

Aaand did you have a look on that? 😇

tilmx:
And another bug: When you copy & paste selected text inside the edit mode, it does copy & paste the _entire element_ instead of just the selected text... And when you hit Cmd+C and then Cmd+V you insert the element inside itself again, which crashes like https://github.com/meetalva/alva/pull/417#issuecomment-386461397

marionebl:
> Just stumbled across one bug: When I drag an element around and drop it on itself – it deletes itself. We already had this bug somewhen in the beginning and it looks like it just came back. If we get that fixed – I’m really lucky! Or is this a topic for a separate PR? @marionebl

Fixed via c64912f30d3087c18ce3d44bc786a7ef5fbad48d

marionebl:
> And another bug: When you copy & paste selected text inside the edit mode, it does copy & paste the entire element instead of just the selected text... And when you hit Cmd+C and then Cmd+V you insert the element inside itself again, which crashes like

To my knowledge copy-pasting is broken on master, too. Let's create an issue + separate PR for this.

tilmx:
>> And another bug: When you copy & paste selected text inside the edit mode, it does copy & paste the entire element instead of just the selected text... And when you hit Cmd+C and then Cmd+V you insert the element inside itself again, which crashes like

>To my knowledge copy-pasting is broken on master, too. Let's create an issue + separate PR for this.

@marionebl Copy-Pasting elements into itself is broken there, too, that’s right. I'll open an issue for that. But could we fix here, that it does not copy the entire element when it is in edit mode? So, e.g. I have "Lorem" as a element name and start editing, select "Lorem", hit Cmd+C, and then hit two times Cmd+V to create "Lorem Lorem", the component crashes.

We had a different thingie like that with the properties (because you can copy/paste there in the inputs as well) and had to something to prevent copying of the entire element when you are inside an input. Do you think we could fix this here, too?

marionebl:
> We had a different thingie like that with the properties (because you can copy/paste there in the inputs as well) and had to something to prevent copying of the entire element when you are inside an input. Do you think we could fix this here, too?

Understood and yes, that's possible. I think the fact we run into this type of thing for the second time shows there is an issue with how copy and pasting is handled currently. That's a bigger but more solid change to the system, alas a new PR.

tilmx:
Okay, then let’s do that. Could you write a new issue for that? That would be awesome!

PR #418, opened 2018-05-03
feat(preview-highlight): update styling of preview highlight

tilmx:
This updates the styling of the preview highlighting, when an element is selected.

## Changes

* Show highlighting all the time, instead of fading it out after 500ms
* Only show the highlight outline, instead of an blue background
* Adapt the outline color to the background

## Tasks

- [x] Restyle Highlighting
- [x] Hide Highlighting when no element is selected
- [x] Update Highlighting position and size when selected element changes (e.g. size)


tilmx:
There is still one bug: If I move the element to another position in the element pane, the highlighting stays still at the old position. I expect the highlighting to move to the same new position as the element. @Palmaswell

tilmx:
Closed as it is already done with #460

Issue #387, opened 2018-04-19
In empty pages, adding the first page element does not work


TheReincarnator:
Create a new page, then open it.
Try to drag-drop a pattern into the elements pane. Nothing happens.

faselbaum:
What is the expected behavior here?
if the element pane is empty should dropping on it set a new root element for that page or do we need to support multiple elements on the root level?

PR #426, opened 2018-05-04
feat: allow users to add new pages

### LINKED ISSUE #24: Create new page
As a user I want to create a new page inside a project

marionebl:
Extend the page list view with a floating button that adds a new page to the list

* Appends page always at end of list
* Selects and focuses new page for editing 

---

Usefulness very limited via https://github.com/meetalva/alva/issues/387, let's tackle this next

Fixes #24 

<details>
<summary>Screencast</summary>

add-page

</details>


tilmx:
@marionebl Nice one! Is there a reason why Undo/Redo Operations do not work for renaming and creating pages? I think we should support that, too. Or what’s your opinion?

tilmx:
Update: As the delete-page function is not built for the moment, an Undo/Redo does not make a lot of sense. The renaming issue with the files will be done in a separate PR.

Issue #422, opened 2018-05-04
List of known Element Pane Bugs


tilmx:
We found out there are a couple of bugs currently in the Element Pane:

- [ ] Copy & Pasting the same page element inside itself
When I copy via Cmd+C an page element, select the same page element and hit Cmd+V, the page element seems to try to insert it in itself and crashes. It should not be possible to hit Cmd+V on the same page element, when it has been copied before.

- [ ] Deleting Root Page Element
When I hit [Backspace] on the root Layout Component, something mysterious happens and Alva is broken. Together with #387 we should discuss, if we can have multiple root elements or set the root element undeletable. We should also think about displaying the root element in a different way (e.g. icon).

- [ ] Hovering is slow
When you have a lot of page elements inside the element pane and hover fast above them from e.g. top to bottom, the performance is really slow and the hover status is late to the mouse pointer.

marionebl:
> When I hit [Backspace] on the root Layout Component, something mysterious happens and Alva is broken. Together with #387 we should discuss, if we can have multiple root elements or set the root element undeletable. We should also think about displaying the root element in a different way (e.g. icon).

Fixed via #437

Issue #23, opened 2017-12-05
Delete page


tilmx:
As a user I want to delete an existing page

kotzendekrabbe:
#83

markusoelhafen:
An existing page can only be deleted on the _page overview_.

First the page needs to be _selected_ (with one click).
There are **3 options to delete a page**:
- _Shortcut_: ⌫
- _Main Menu_: Edit --> 'Delete page'
- _Right Click_: Context menu -> 'Delete page'

Before the deleting the page, an **alert** will show and ask the user if he really wants to delete the page.

**Example Text:**
Are you sure you want to delete "[projectname]"?

**Alert Options:**
"Delete" -> Finally deletes the page
"Cancel" -> cancel action

**Flow:**
![deletepage](https://user-images.githubusercontent.com/10544994/37663376-5fbcb348-2c59-11e8-9995-d4bb89da429e.png)

PR #437, opened 2018-05-07
 refactor: unify page/page-element operations as store methods 

### LINKED ISSUE #240: refactor menu.ts
We should think about to refactor the menu.ts file.

I would suggest to create a new folder called `menu`and move the current `menu.ts` there (maybe rename it `index.ts`). Then create new files like `create_styleguide.ts` and move the logic from the  menu there.

### LINKED ISSUE #423: Refactor global keyboard handling
## Situation

As uncovered in #417 the `shortcut` keys of `menu` items in Alva's system menu interfere in unexpected ways with default behavior of common HTML-builtins, such as `<input/>`.

I haven't analyzed this fully, but my current assumption is that there is no way prevent an items `action` fire for a configured keyboard shortcut, even if executing `e.stopPropagation` on a event during its bubbling phase.

## Proposal

We should move the global shortcut listeners into the context of the renderer application. This way we get fine-grainded control over propagation, prevent defaulting etc.

## Related issues

* #422 
* #148

marionebl:
* Fixes #240 
* Fixes issue where elements pasted on themselves disappeared (compare: #422)
* Fixes #423 
* Prepares PR for #23


marionebl:
>  When duplicate the element via Cmd+D, not the newly inserted element is selected, but the previous one I copied stays selected. The new element should be always selected.

Adressed via https://github.com/meetalva/alva/pull/437/commits/e6995d8c21b8e934b3ef3479528544859d986426

tilmx:
Aaaaand: When I create a new page now the root element is not added any more :-(

marionebl:
> When I'm inside the edit mode an page element – the context menu on right-click is still there and suggests that copy/paste would work – but it doesn’t. Can we either disable the context menu in this case or add the functionality of copying/pasting/cutting the selected text (instead of the element)?

Addressed by adapting the context menu when the clicked item is in editing mode. `paste` is disabled due to buggy behavior I didn't want to invest the time to fully analyze.

PR #447, opened 2018-05-08
WIP: feat(migrator): add migrator logic for converting old files

LKuechler:


marionebl:
Hey @LKuechler, let's pick this up and adopt it to the `.alva` file structure. 

We might want to implement it roughly like this:

### In [models/project](https://github.com/meetalva/alva/blob/master/src/model/project.ts)

I recommend we anchor the new mechanism on the `Project` model, because that is the entity that is persisted to disk. `Project` instances expose the `toDisk` method, which emits a serializable object representing the instance suitable for saving to the persistence.

1. Introduce a `.version` field in `Project` that holds the current semver version string
2. Introduce a method `Project#getMigrations`, which encapsulates semver comparison and later on more complexity we might want to cover:

   ```ts
   getMigrations(project: string, semanticRange: string): Migration[];
   ```
I'd begin with comparing the received version string `a` with a version `b` kept as private static field on `Project` and return all migrations that are `b < && > a`. Which precludes migrations having a `.version` field, too.

### In [persistence/persistence](https://github.com/meetalva/alva/blob/master/src/persistence/persistence.ts#L60)

1. When reading from disk, pass in `T` to `Persistence.read`, e.g.:

   ```ts
   public static read<T>(path: string, context: { model: T }): Promise<PersistenceReadResult<T>>;
   ```

2. Use `context.model#getMigrations` to obtain required migrations on the `SerializedProject` data structure and execute them in series on the input data. 

What do you think?

marionebl:
Closing as discussed with @LKuechler in person.

PR #452, opened 2018-05-11
refactor(export): use message system for all exporters

LKuechler:


src/electron/main.ts
```diff
@@ -83,6 +84,7 @@ async function createWindow(): Promise<void> {
 
 		// Handle messages that require
 		// access to system / fs
+		// tslint:disable-next-line:cyclomatic-complexity
```
LKuechler: do we want to move the code inside the cases to seperate functions instead of disabling the cyclomatic-complexity check?

LKuechler: @marionebl and i decided that it does not make sense to move the code in to different function.
This would not make it more clear to read and the current function will not get less complex because of it.

PR #455, opened 2018-05-18
WIP: Feature/drag of image files on elementlist

### LINKED ISSUE #28: Drag-and-drop support for the placeholder component
- [ ] I should be able to drag a PNG or JPG from my local folder into the page element layers.

LKuechler:
…y on the element-list
Fixes #28

### DIFF
--- a/src/container/element-list/element-list.tsx
+++ b/src/container/element-list/element-list.tsx
@@ -212,39 +212,7 @@ export class ElementList extends React.Component {
 			return;
 		}
 
-		const getDropIndex = () => {
-			if (!isSiblingDrop) {
-				return dropTargetContent.getElements().length;
-			}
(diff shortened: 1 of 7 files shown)


tilmx:
Works fine for me!

tilmx:
But can we only allow .png, .jpg and .svg to drop? Otherwise I can drag anything like an .zip or .app into Alva, but the placeholder doesn’t render any more.

tilmx:
@LKuechler It does not indicate the drop zones any more when I drag an PNG or any other file onto the element pane. Before the latest change, you could see wether you drop it as a children or between. Can you reproduce that or should I attach a video?

marionebl:
Closing as discussed with @LKuechler in person.

PR #461, opened 2018-05-24
feat: select items from preview

marionebl:
Fixes #130

PR #462, opened 2018-05-24
feat: select elements from preview

### LINKED ISSUE #130: Select Element on click
When clicking on a page element in the preview pane, it should be selected in the pattern pane.

marionebl:
Closes #130 
Replaces #461

### DIFF
--- a/src/container/element-list/element-wrapper.tsx
+++ b/src/container/element-list/element-wrapper.tsx
@@ -1,11 +1,10 @@
-import { Element, ElementAnchors } from '../../components';
+import { Element } from '../../components';
 import * as React from 'react';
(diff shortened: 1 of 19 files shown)


src/container/element-list/element-wrapper.tsx
```diff
@@ -133,7 +119,7 @@ export class ElementWrapper extends React.Component<ElementWrapperProps, Element
 	}
 }
 
-function above(node: EventTarget, selector: string): HTMLElement | null {
+/* function above(node: EventTarget, selector: string): HTMLElement | null {
```
tilmx: Should we remove that code? @marionebl

tilmx:
Runs smooth for me now!

PR #463, opened 2018-05-25
fix: do not connect to ws server from html export

marionebl:
Fixes a bug where the HTML export would not work because connection to the Alva Websocket server was not possible


tilmx:
Works for me!

PR #483, opened 2018-05-30
build(distribution): remove designkit from alva build steps

LKuechler:


scripts/dist-preprocess.js
```diff
@@ -1,26 +1,9 @@
...
-	npm run build &&
-	rm -rf node_modules &&
-	npm i --production
-`, {
-		cwd: 'build/designkit'
-	}
-);
-
-console.info('Distrubution preperation finished.');
+// console.info('Distribution preparation finished.');
```
tilmx: Remove the line?

LKuechler: good catch. I removed it :)

PR #496, opened 2018-05-31
feat: add description to property

tilmx:


src/components/property-items/string-item/index.tsx
```diff
@@ -53,18 +60,21 @@ const StyledInput = styled.input`
...
+				<PropertyLabel label={label} />
+				<StyledInput
+					onChange={onChange}
+					onBlur={onBlur}
+					type="text"
+					value={value || ''}
+					placeholder="…"
+				/>
+			</StyledContainer>
+			{description !== '' && <PropertyDescription description={description || ''} />}
```
tilmx: The description is still shown as empty element – but I only want to show it when it is not ''. Can you help me with that @marionebl?

src/analyzer/typescript-react-analyzer/property-analyzer/property-analyzer.ts
```diff
@@ -128,6 +128,7 @@ function createArrayProperty(
 		return {
 			contextId: args.symbol.name,
 			defaultValue: [],
+			description: args.symbol.name,
```
marionebl: I think we should use an empty string here: `''`

PR #497, opened 2018-05-31
fix(exporter): trigger a second png export if the first does not retu…

LKuechler:
…rn a valid value


LKuechler:
This is not a very good solution but the event for the finished loading fires to early and there seem to be a bug in electron with a timeout combined with a webview

PR #519, opened 2018-06-05
fix: disable text select on splash screen

tilmx:
Disable user-select of text on Splash Screen, so it feels like a desktop app.

### DIFF
--- a/src/components/splash-screen/index.tsx
+++ b/src/components/splash-screen/index.tsx
@@ -11,6 +11,8 @@ const StyledSplashScreen = styled.div`
 	justify-content: center;
 	padding: 0 ${PageInset.XL}px;
 	transform: translateY(-54px);
+	user-select: none;
+	cursor: default;
 `;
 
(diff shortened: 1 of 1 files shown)


src/components/headline/index.tsx
```diff
@@ -21,6 +21,7 @@ const StyledHeadline = styled.div`
 	margin-top: 0;
 	font-family: ${fonts().NORMAL_FONT};
 	font-weight: 500;
+	user-select: none;
```
marionebl: I am not a huge fan of disableing text selection for entire typography styles. If we positively need to block it somewhere we should do it on that specific screen.

marionebl:
Hey there, let's discuss this again - my opinion on blocking user selection is: Don't - having everything selectable may well be a super power of web based interfaces compared to native ones. Of course there are exceptions (e.g. when selection interferes with primary interaction).

tilmx:
Heyhey! Yes, let’s discuss this tomorrow @marionebl

PR #522, opened 2018-06-05
feat: add bug report button

tilmx:
Adds a bug report button in the chrome that leads to the Github Issues page with bugs. This might be an easy approach for users during the Alpha/Beta phase to immediately get the right location for sending bugs.

Should be squashed into `feat: add bug report button`


src/container/chrome/chrome-container.tsx
```diff
@@ -70,6 +71,14 @@ export const ChromeContainer = MobxReact.inject('store')(
 						title={project ? project.getName() : 'Alva'}
 					/>
 				)}
+				<BugReport
+					title="Found a bug?"
+					onClick={() => {
+						Electron.shell.openExternal(
```
marionebl: We are in the process of moving all `Electron` dependencies to the main process. We should move this call to `electron/main.js` and send an event to trigger it instead

tilmx: Sounds like a good idea, but how would I do that @marionebl?

tilmx: Got an idea. Trying it out now.

tilmx:
Moved it to the main process @marionebl

src/components/bug-report/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+import styled from 'styled-components';
+
+export interface BugReportProps {
+	onClick?: React.MouseEventHandler<HTMLElement>;
+	title: string;
+}
+
+const StyledBugReport = styled.div`
+	justify-self: right;
+	margin-right: -${getSpace(SpaceSize.XXL) * 3 - getSpace(SpaceSize.L)}px; // align to top right corner
```
LKuechler: The padding should be removed in the `chrome` component

src/container/chrome/chrome-container.tsx
```diff
@@ -70,6 +73,17 @@ export const ChromeContainer = MobxReact.inject('store')(
 						title={project ? project.getName() : 'Alva'}
 					/>
 				)}
+				<BugReport
+					title="Found a bug?"
+					onClick={() => {
+						console.log('chrome-container started');
```
LKuechler: This is just debug code right? ^^

src/components/bug-report/index.tsx
tilmx: Yes, but the padding leads to the fact, that the "< Pages" indicator is under the 3 mac close icons. Or if we only set the right padding differently, the center title is not properly aligned (a bit too right). That’s why I chose that solution.

src/container/chrome/chrome-container.tsx
tilmx: Yes. Good spot!

PR #526, opened 2018-06-06
WIP feat(export-sketch): trigger info message before sketch export

LKuechler:
The info message gives some information on how to use the sketch export with sketch.


tilmx:
We’ll have to figure out how do deal with links within the alert window – thought about integrating an dialog instead and having an option to open a link instead (and handle that on callbacks). I’ll have a look on that today evening.

tilmx:
Closed in favor of #544

PR #531, opened 2018-06-06
fix(typescript-react-analyzer): analyze number enums correctly

tilmx:
Number Enums were parsed as Strings. They are now parsed as numbers via parseInt.
The headline component from the designkit should now have working enums again.

Should be squashed into `fix: analyze number enums correctly`


LKuechler:
@tilmx 
in your target squash commit you can also add a scope: `fix(typescript-react-analyzer): analyze number enums correctly`

Issue #534, opened 2018-06-07
Opening of a alva file can be reverted with `undo`


LKuechler:
**Describe the bug**
When Alva is opened with clicking on a Alva project file this can be reverted with cmd+z.

**To Reproduce**
1. Make sure that Alva is closed
2. Open a Alva project
3. Click `Edit -> Undo` or `cmd+z`
4. Now the window should be white

**Expected behavior**
The history should not include the opening of project files

**Screenshots**
![alva-bug-undo-opening](https://user-images.githubusercontent.com/2543933/41120762-cf5582ae-6a96-11e8-9990-9daacf93fc28.gif)

**Device (please complete the following information):**
 - macOS 10.13.4
 - Current master (commit hash: 9c35fe53983)

LKuechler:
fixed with #536

PR #536, opened 2018-06-07
fix: stop committing openings of files in store

barinali:
Aimed for #534, openings of an Alva file can be reverted with `undo`.

Therefore, I have removed `store.commit()` from `ServerMessageType.OpenFileResponse` and `ServerMessageType.CreateNewFileResponse` in order to ignore these kind of changes for history.

### DIFF
--- a/src/electron/renderer.tsx
+++ b/src/electron/renderer.tsx
@@ -71,8 +71,6 @@ Sender.receive(message => {
 					type: ServerMessageType.CheckLibraryRequest
 				});
 			}
-
-			store.commit();
 			break;
 		}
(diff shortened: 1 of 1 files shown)


tilmx:
Sounds good! Thanks for contributing!

tilmx:
Hey @barinali and @LKuechler – I just tried it out and run into a little problem – the first action after creating a file/project can’t be undone now. E.g. – I open Alva, Create a project, add a text to the elements and Hit Cmd+Z, the text does not remove any more.

Can we somehow fix this?

barinali:
@tilmx I'll look at it and try to come up with a solution.

PR #543, opened 2018-06-10
feat: show error if library cannot be connected

tilmx:
If the user tries to connect a library folder that does not work, we should give him a warning about that.

should be squashed into `feat: show error if library cannot be connected`

### DIFF
--- a/src/electron/main.ts
+++ b/src/electron/main.ts
@@ -305,14 +305,34 @@ const userStore = new ElectronStore();
 					return;
 				}
 
-				const analysis = await Analyzer.analyze(connection.path, {
-					getGobalEnumOptionId: (patternId, contextId) =>
-						library.assignEnumOptionId(patternId, contextId),
-					getGlobalPatternId: contextId => library.assignPatternId(contextId),
-					getGlobalPropertyId: (patternId, contextId) =>
-						library.assignPropertyId(patternId, contextId),
-					getGlobalSlotId: (patternId, contextId) => library.assignSlotId(patternId, contextId)
-				});
+				let analysis;
+
+				try {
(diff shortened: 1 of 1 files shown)


src/electron/main.ts
```diff
@@ -305,14 +305,34 @@ const userStore = new ElectronStore();
...
 
-				const analysis = await Analyzer.analyze(connection.path, {
-					getGobalEnumOptionId: (patternId, contextId) =>
-						library.assignEnumOptionId(patternId, contextId),
-					getGlobalPatternId: contextId => library.assignPatternId(contextId),
-					getGlobalPropertyId: (patternId, contextId) =>
-						library.assignPropertyId(patternId, contextId),
-					getGlobalSlotId: (patternId, contextId) => library.assignSlotId(patternId, contextId)
-				});
+				let analysis;
```
marionebl: Let's avoid the `let` binding by defining a `getAnalysis` function that returns a `AnalysisResult`: 

```ts
type AnalysisResult =
  | AnalysisError
  | AnalysisSuccess;

enum AnalysisState {
   Error,
   Success
}

interface AnalysisError {
   state: AnalysisState.Error;
   error: Error;
}

interface AnalysisSuccess {
   state: AnalysisState.Success;
   analysis: Analysis;
}

const getAnalysis = async (path: string; context: AnalysisContext): AnalysisResult => {
  try {
    return { state: AnalysisState.Sucess, analysis: Analyzer.analyze(path, context) };
  } catch (error) {
    return { state: AnalysisState.Error, error };
  }
};
```

tilmx:
@marionebl I changed it to try both the analysis and the Sender.Send to avoid any `let: analysis;` thingies.

PR #553, opened 2018-06-11
feat(editable-title): make page title editable

Palmaswell:


marionebl:
Looks good, just found an older bug that might be more obvious now.

1. Edit the name of a page
2. Make it longer than e.g.: 15x`a`
3. `text-overflow: ellipsis` strikes back

screen shot 2018-06-12 at 23 40 26

src/components/editable-title/index.tsx
```diff
@@ -0,0 +1,149 @@
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import styled from 'styled-components';
+
+import { Color } from '../colors';
+import { CopySize } from '../copy';
+import { getSpace, SpaceSize } from '../space';
+
+export enum EditState {
```
marionebl: Should be `EditableTitleState`

src/components/editable-title/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+	secondary?: boolean;
+}
+
+interface StyledInputProps {
+	fontSize?: CopySize;
+	secondary?: boolean;
+}
+
+const StyledTitle = (props: StyledEditableTitleProps): JSX.Element => {
+	const Strong = styled.strong`
```
marionebl: Remove `StyledTitle` and use `Strong` => `StyledStrong` directly.

You can add `data-title` via

```
const StyledStrong = styled.strong.attrs({ "data-title": true })``
```

src/components/editable-title/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+import { CopySize } from '../copy';
+import { getSpace, SpaceSize } from '../space';
+
+export enum EditState {
+	Editable = 'Editable',
+	Editing = 'Editing'
+}
+
+export interface EditableTitleProps {
+	fontSize?: CopySize;
```
marionebl: Could `fontSize` be bound to `secondary`?

src/components/editable-title/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+	focused: boolean;
+	name: string;
+	nameState: EditState;
+	secondary?: boolean;
```
marionebl: Make this an enum, make it required

```
export EditableTitleOrder {
   Primary,
   Secondary
}
```

src/components/editable-title/index.tsx
Palmaswell: done

src/components/editable-title/index.tsx
Palmaswell: That totally makes sense. Removed the fontSize prop

src/components/editable-title/index.tsx
Palmaswell: Included the new enumerable to categorized the different EditableTitle component. Which I agree to be a good idea to give a much clearer idea of the component props intent :)

Palmaswell: I will take care now of abstracting the `EditableTitleState.Editing` and `EditableTitleState.Editable` store to the container component own state in order to solve the problem mentioned below by @tilmx.

tilmx:
Looks fine to me! Just to things:
- [ ] The Page Tiles should have 12px font-size, not 15px (the chrome bar should stay at 15px as it is now)
- [ ] Blurring the title input should save the edited title, not abort it

Issue #555, opened 2018-06-12
First Action cannot be undone


tilmx:
Issue assigned to @barinali 

**Describe the bug**
The first action after opening or creating a project cannot be undone.

**To Reproduce**
1. Open or create a new project
2. Do something (e.g. add a new element to the element list)
3. Try Cmd+Z

**Expected behavior**
The first action should be undo-able, as well.

barinali:
Hey @tilmx,

Sorry that I couldn't come back with a solution yet because of lacking time. But I have an intention to fix this since I somehow caused that. Is it possible to assign this issue to me?

tilmx:
Hey @barinali 
yes, of course! I’ll assign it to you. Let us know if you need anything!
And thanks for contributing!

barinali:
I have opened a pull request for this, #556.

While working with the history, I noticed that we can apply 'undo' command to go back from "Pages" view to "PageDetail" view and vice versa. 🤔 I do not think that should be possible. So I would suggest removing `store.commit()` from `ViewStore.setActiveAppView`. Also, if we do that, this will require us to manage separate histories for each `AlvaView`. Because while being on "Pages", we shouldn't be able to undo/redo actions for "PageDetail" view (like undo-ing adding a component) and so on.

Therefore, I guess file history management/actions (undo, redo) need some work.

tilmx:
Hey @barinali 
Sounds good, thank you very much! Your note makes totally sense, we are currently in the process of bringing the page view into the page detail view (#554) – to avoid problems as you noticed and to speed up page switching.

tilmx:
Closed with #556

marionebl:
Closed via landing #556, thanks @barinali!

> While working with the history, I noticed that we can apply 'undo' command to go back from "Pages" view to "PageDetail" view and vice versa. 🤔

Impressive you catched the quirkiness of the undo/redo system on the first try! We went back and forth about the undo/redo feature internally. @tilmx decided on a flow where every undo/redo is executable regardless of current view but the app shows the relevant view when performing the operation.

barinali:
@tilmx @marionebl you are welcome! I am happy to help as long as I can.

I can't wait to see #554 in! 👍

PR #556, opened 2018-06-12
fix: make first action undo-able

barinali:
Aimed for #555, the first action cannot be undone.

This issue helped me understand how the history works in the store. :)

So.. I think we don't have to keep/commit every change in the store. Instead, to make the first action undo-able, I have committed the state whenever a file is opened/created. Because it allows us to know the previous state of the current file that we can go back. Besides that, I have stopped committing the state in `ServerMessageType.StartApp`. Because that was having a role in making #534 possible.


marionebl:
Thank you 👍

PR #557, opened 2018-06-13
chore: use patternplate cover as website mechanism

marionebl:
This PR demonstrates how we could use [patternplate cover](https://patternplate.github.io/doc/docs/guides/cover?guides-enabled=true) to create the meetalva.io website.

## Steps to test

* `yarn patternplate`
* Access http://localhost:1337 for the website
* Access http://localhost:1337/#be-part-of-the-future-of-design for a link leading to the underlying patternplate interface
* Access http://localhost:1337/?reload=true for auto-reloading capabilities

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -72,9 +72,9 @@
       }
     },
     "@babel/runtime": {
-      "version": "7.0.0-beta.49",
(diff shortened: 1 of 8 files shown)


tilmx:
As discussed with Mario: The idea is cool, but at the moment there is no appropriate content for the cover. With our new website for the Beta Launch we’ll add a Getting Started page with more information on how to use Alva. That website might be a good case to take it as a cover!

PR #558, opened 2018-06-13
feat: add shortcut to switch pages

markusoelhafen:
For easier page switches I added two shortcuts.

## Content
* Add shortcuts for next and previous page

## Todo
- [x] Add shortcuts to menu
- [x] Only enable "next page" when there's a page
- [x] Only enable "previous page" when there's a page

### DIFF
--- a/src/electron/create-menu.ts
+++ b/src/electron/create-menu.ts
@@ -395,6 +395,7 @@ export function createMenu(ctx: MenuContext): void {
 				{
 					label: 'Next Page',
 					accelerator: 'CmdOrCtrl+Alt+Right',
+					enabled: ctx.store.getLastPage() !== ctx.store.getCurrentPage(),
 					click: () => {
 						ctx.store.setNextPage();
 					}
--- a/src/electron/create-menu.ts
+++ b/src/electron/create-menu.ts
@@ -385,6 +385,20 @@ export function createMenu(ctx: MenuContext): void {
 					type: 'separator'
 				},
 				{
+					label: 'Next Page',
+					accelerator: 'CmdOrCtrl+Alt+Left',
+					click: () => {
+						ctx.store.setPreviousPage();
+					}
+				},
+				{
+					label: 'Next Page',
+					accelerator: 'CmdOrCtrl+Alt+Right',
+					click: () => {
+						ctx.store.setNextPage();
+					}
+				},
+				{
 					label: '&Show Elements & Components',
 					type: 'checkbox',
 					checked: true,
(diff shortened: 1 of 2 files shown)


src/electron/create-menu.ts
```diff
@@ -384,6 +384,22 @@ export function createMenu(ctx: MenuContext): void {
 				{
 					type: 'separator'
 				},
+				{
+					label: 'Previous Page',
+					accelerator: 'CmdOrCtrl+Alt+Left',
+					enabled: ctx.store.getFirstPage() !== ctx.store.getCurrentPage(),
```
marionebl: This will evaluate to true if both `currentPage` and `firstPage` are undefined, e.g. when there are no pages at all. 
A more robust approach could be to try and get the previous page and check if it is not undefined, e.g.:

```
enabled: typeof ctx.store.getPreviousPage() !== 'undefined'
```

src/electron/create-menu.ts
```diff
@@ -384,6 +384,22 @@ export function createMenu(ctx: MenuContext): void {
...
+					click: () => {
+						ctx.store.setPreviousPage();
+					}
+				},
+				{
+					label: 'Next Page',
+					accelerator: 'CmdOrCtrl+Alt+Right',
+					enabled: ctx.store.getLastPage() !== ctx.store.getCurrentPage(),
```
marionebl: See comment on https://github.com/meetalva/alva/pull/558/files#diff-d3a5c49b9ddc2f6f39d233553a4f9c7dR390

src/store/view-store.ts
```diff
@@ -601,6 +601,11 @@ export class ViewStore {
 		return this.project.getElements().find(element => element.getHighlighted());
 	}
 
+	@Mobx.action
+	public getFirstPage(): Model.Page {
```
marionebl: Let's remove `getFirstPage` and `getLastPage`, related to the comments on https://github.com/meetalva/alva/pull/558/files#diff-d3a5c49b9ddc2f6f39d233553a4f9c7dR398, https://github.com/meetalva/alva/pull/558/files#diff-d3a5c49b9ddc2f6f39d233553a4f9c7dR390

Let's introduce `getNextPage` and `getPreviousPage` instead. Using those we could do:

```ts
const nextPage = store.getNextPage();

if (nextPage) {
  store.setActivepage(nextPage);
}
```

markusoelhafen:
@marionebl thx for the feedback! fixed the store and the menu :)

marionebl:
Thank you <3

PR #560, opened 2018-06-13
feat: add tab switch to hide pages and sidebar

tilmx:
Add a tab switch to hide both the pages or the elements & library:

bildschirmfoto 2018-06-13 um 21 39 19


markusoelhafen:
I don't see the necessity for the underline to highlight the active icon in addition to the color

tilmx:
It makes more sense when this pattern is used on the right side as a sidebar switch to switch between Properties and Project Settings, doesn’t it?

bildschirmfoto 2018-06-13 um 21 45 00

markusoelhafen:
Also, if there are options to toggle pages and the elementpane, there should also be an option to hide the property pane

markusoelhafen:
@tilmx if it's used as an actual tab I agree, but if it's used as an checkbox I'd disagree.
In tabs, the line might helps to see the switch between two views in the same pane.

tilmx:
I removed it.

src/components/tab-switch/tab-switch.tsx
```diff
@@ -0,0 +1,40 @@
+import { Color } from '../colors';
+import { Copy } from '../copy';
+import { Icon, IconName, IconSize } from '../icons';
+import * as React from 'react';
+import { getSpace, SpaceSize } from '../space';
+import styled from 'styled-components';
+
+export interface TabSwitchProps {
+	active?: boolean;
```
marionebl: Let's use an enum instead:

```
enum TabSwitchState {
  Active,
  Default
}
```

src/electron/main.ts
```diff
@@ -508,7 +507,7 @@ async function createWindow(): Promise<void> {
 		minWidth: 780,
 		minHeight: 380,
 		titleBarStyle: 'hiddenInset',
-		backgroundColor: Color.Grey97,
+		backgroundColor: '#f7f7f7',
```
marionebl: Some special reason for this one?

tilmx: Color.Grey97 didn’t work (I guess because electron only allows hex values)

src/components/tab-switch/tab-switch.tsx
tilmx: Did that!

src/electron/main.ts
marionebl: Ah, so lets use the `color` module here to convert from rgb to hex then.

PR #561, opened 2018-06-13
 feat: update styling of event handler property

tilmx:
Adds a new styling to the event handler property.
Those properties now have a box styling to make it clear that the content is connected:

bildschirmfoto 2018-06-13 um 22 18 26


src/components/property-box/index.tsx
```diff
@@ -0,0 +1,31 @@
+import { Color } from '../colors';
+import { Copy } from '../copy';
+import { Headline } from '../headline';
+import * as React from 'react';
+import { getSpace, SpaceSize } from '../space';
+import styled from 'styled-components';
+
+export interface PropertyBoxProps {
+	headline?: string;
```
marionebl: Let's make `headline` and `copy` required

tilmx: Fixed that!

src/components/property-box/pattern.json
```diff
@@ -0,0 +1,7 @@
+{
+	"name": "box-item",
```
marionebl: Should be `property-box`

tilmx: Done!

PR #570, opened 2018-06-16
feat: introduce layout switch

tilmx:
I introduced a new layout switch inspired from [iA Writer](https://ia.net/de/writer) where you can hide or show all sidebars at once or specify which one should be displayed and which one not.

Btw, let’s try out Feather Icons (https://github.com/feathericons/feather) as a base for our icons.


tsconfig.json
```diff
@@ -15,19 +15,11 @@
 		"isolatedModules": false,
 		"jsx": "react",
 		"module": "commonjs",
-		"plugins": [
```
marionebl: Any reason to remove this for this PR?

tilmx: Didn’t remove it entirely, was just a duplicate...

PR #580, opened 2018-07-02
fix: allow sidebars to scroll

### LINKED ISSUE #572: Page Pane does not scroll
**Describe the bug**
If there are too many pages to fit into the viewport in the page pane, you can’t scroll inside the page pane. That leads to the problem that you can’t access pages any more that are outside the initial view.

**To Reproduce**
1. Create/Open a Alva Project
2. Add a lot of pages, so there are more pages than space for it
3. Try to scroll to the last one

**Expected behavior**
I should be able to scroll down the pages!

**Additional context**
We should set an `overflow: auto` to the page pane.

tilmx:
If a sidebar has too much content for the viewport, you can scroll down now.

Fixes #572


barinali:
LGTM. 😬

Issue #571, opened 2018-06-18
Double-Clicking on Chrome Buttons leads to Maximize/Unmaximize


tilmx:
**Describe the bug**
Double-clicking on chrome buttons leads the same event as double-clicking the entire chrome bar (maximizing or un-maximizing the window).

**To Reproduce**
1. Open Alva Project
2. Double-Click on Layout Switch in Chrome or left or right Page-Switch-Button in Chrome

**Expected behavior**
When double-clicking a button or switch on the chrome bar, the windows should not be maximized or un-maximized. That should only happen, if you double-click on the rest of the chrome area. 

**Screenshots**

**Additional context**
We could test double-click events on e.target, if the user hits the chrome or a specific button on it.

marionebl:
Fixed via #584 
Thanks again @barinali !

PR #584, opened 2018-07-11
fix: disable doubleClick for chrome elements

barinali:
I have added `doubleClick` event along with `event.stopPropagation` to elements in `<Chrome />` not to trigger `<Chrome />`'s `doubleClick` event. This pull request should resolve issue #571.


marionebl:
Hey @barinali, thanks for tackling this! I'll have a look at your changes as soon as a mid-sized refactoring of our data models went through. 👍

PR #585, opened 2018-07-12
feat: close preview server to other network devices

### LINKED ISSUE #573: Alva creates global webserver for preview, not only local
**Describe the bug**
The preview uses a web server via an iFrame for all the preview stuff. That server does not only serve on localhost, but registers as a webserver in the entire network. That leads to the problem that everybody in the same network can access the preview and highlight or click on elements.

**To Reproduce**
1. Create or open an Alva project
2. Take another device that is on the same network
3. Try `[YOUR ALVA DEVICE IP]:1879/preview.html`
4. Do things

**Expected behavior**
Alva should only create the webserver running on localhost, not on the entire network.

**Screenshots**

**Device (please complete the following information):**
 - OS & Version: macOS 10.13.4
 - Version of Alva: master-branch on 42da585

barinali:
Fixes #573 

This pull request contains an application middleware to check and interrupt the requests depending on the remote address and the host/server address is equal to each other excluding ports.

### DIFF
--- a/src/electron/server.ts
+++ b/src/electron/server.ts
@@ -127,6 +127,23 @@ export async function createServer(opts: ServerOptions): Promise<EventEmitter> {
 		ws.send(JSON.stringify(state));
 	});
 
+	app.use((req, res, next) => {
+		// get IPv4 ip out of IPv4-mapped IPv6 address
+		const remoteAddress = ((req.connection.remoteAddress as string) || '').split(':').pop();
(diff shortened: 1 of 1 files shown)


marionebl:
Thanks for your contribution. #573 will be solved when the changes from #586 land, see
https://github.com/meetalva/alva/pull/586/files#diff-bad55d25e99d158455abee5f597be4e4R82

PR #600, opened 2018-09-11
fix: add element contents for unoccupied pattern slots

marionebl:
Adresses a bug where adding a pattern slot to a pattern with
callsites would not update the appropriate elements inside Alva.

Steps to reproduce

1. Create new file
2. Connect DesignKit
3. Add element with Button pattern
4. Remove children prop from Button interface
5. Update the library inside Alva
6. Readd children prop to Button interface
7. Update the library inside Alva again
7a) Before fix: The button still does not accept children
7b) After fix: The button accepts children again

### DIFF
--- a/src/model/element/element-content.ts
+++ b/src/model/element/element-content.ts
@@ -80,6 +80,20 @@ export class ElementContent {
 		);
 	}
 
+	public static fromSlot(slot: PatternSlot, context: ElementContentContext): ElementContent {
+		return new ElementContent(
+			{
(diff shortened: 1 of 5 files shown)


marionebl:
>  i just found another bug that gets introduced with the changes.
> 1. Use Teaser Component
> 2. See two slots (Button, Headline)
> 3. Update All Libraries
> 4. See two more slots
> 5. Repeat from 3.

Good catch, thanks!

PR #605, opened 2018-09-14
feat: create projects without explicit path

marionebl:
This changes the basic behavior to ask for a save path on quit / window close / save for draft projects instead of asking up front. 

Draft projects are saved in the application support files when first created and marked as draft. When one of the following events occur for a draft project, the user is asked for a path to save to:

* `Cmd+S` (draft projects)
* `Cmd+Shift+S` (all projects)
* Any dialog schedules a save transaction

Additionally, when trying to close a window with a draft project active the user is asked wether the current project should be saved, discarded or the close canceled.

<details>
<summary>Dialog</summary>
screen shot 2018-09-14 at 22 47 40
</details>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2336,6 +2336,16 @@
       "integrity": "sha512-BLM56aPo9vLLFVa8+/+pJLnrZ7QGGTVHWsCwieAWT9o9K8UeGaQbzZbGoabWLOo2ksBCztoXdqBZBplqLDDCSg==",
       "dev": true
     },
+    "are-we-there-yet": {
+      "version": "1.1.5",
+      "resolved": "https://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.5.tgz",
+      "integrity": "sha512-5hYdAkZlcG8tOLujVDTgCT+uPX0VnpAH28gWsLfzpXYm7wP6mp5Q/gYyR7YQ0cKVJcXJnl3j2kpBan13PtQf6w==",
+      "optional": true,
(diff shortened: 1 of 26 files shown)


tilmx:
@marionebl Looks great! I stumbled across just two thingies:

- [ ] Approx a second after creating a new file, the name jumps from "New Project" to a long uuid. See attached image:
alva

- [ ] When hitting cmd+s, Alva shows the save dialog bound on the same window (not as a separate view). but when I try to close Alva, click on the upcoming "Do you want to save?"-Dialog on "Save" – it is a separate window. Can we bind this save dialog also to the main window?

marionebl:
> When hitting cmd+s, Alva shows the save dialog bound on the same window (not as a separate view). but when I try to close Alva, click on the upcoming "Do you want to save?"-Dialog on "Save" – it is a separate window. Can we bind this save dialog also to the main window?

I am pretty sure this is caused be the "Do you want to save" dialog box still being active when we init the save select dialog, causing it to detach itself from the window. Adding a timeout before triggering the save transaction might solve this, but I'd rather live with the current state.

tilmx:
> I am pretty sure this is caused be the "Do you want to save" dialog box still being active when we init the save select dialog, causing it to detach itself from the window. Adding a timeout before triggering the save transaction might solve this, but I'd rather live with the current state.

Okay, let’s do that!

marionebl:
> Approx a second after creating a new file, the name jumps from "New Project" to a long uuid. See attached image:

Fixed

PR #607, opened 2018-09-15
WIP feat: ux improvements

tilmx:


src/container/element-list/element-list.tsx
```diff
@@ -403,6 +406,15 @@ export class ElementList extends React.Component {
 			>
 				<Components.Element.ElementChildren>
 					{childContent ? <ElementContentContainer content={childContent} /> : null}
+					{hasChildren ? (
```
marionebl: Should be simplified to 

```
!hasChildren && (<Components.EmptyState ... />
```

src/components/add-button/index.tsx
```diff
@@ -6,41 +6,47 @@ import { Plus } from 'react-feather';
...
 	position: relative;
 	box-sizing: border-box;
 	height: 60px;
 	width: 100%;
-	border: 1px solid ${Color.Grey80};
+	border: 1px solid ${Color.Grey90};
 	border-radius: 6px;
 	background-color: transparent;
-	margin: ${getSpace(SpaceSize.XS)}px;
+	margin: ${(props: AddButtonProps) => (props.margin ? `${getSpace(SpaceSize.XS)}px` : '0')};
```
marionebl: Could the decision wether `margin` should be used placed outside the component via `<Space/>`?

src/components/add-button/index.tsx
```diff
@@ -6,41 +6,47 @@ import { Plus } from 'react-feather';
...
 `;
 
 const StyledIcon = styled(Plus)`
 	margin-right: ${getSpace(SpaceSize.XS)}px;
 `;
 
-export const AddPageButton: React.SFC<AddPageButtonProps> = props => (
-	<StyledAddPageButton onClick={props.onClick}>
+export const AddButton: React.SFC<AddButtonProps> = props => (
+	<StyledAddButton margin={props.margin} text={props.text} onClick={props.onClick}>
```
marionebl: Remove the text prop here and use a distinct `StyledAddButtonInterface` instead

src/components/chrome-button/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+	}
+`;
+
+const StyledIcon = styled.div`
+	margin-right: 6px;
+`;
+
+export const ChromeButton: React.StatelessComponent<ChromeButtonProps> = props => (
+	<StyledChromeButton {...props}>
+		{props.icon ? <StyledIcon>{props.icon}</StyledIcon> : ''}
```
marionebl: Simplify to `props.icon && <StyledIcon .../>`

src/components/copy/index.tsx
```diff
@@ -21,6 +21,8 @@ export enum CopySize {
 const StyledCopy = styled.p`
 	margin: 0;
 	line-height: 1.5;
+	user-select: none;
```
marionebl: :/ Sure?

src/components/headline/index.tsx
```diff
@@ -21,6 +21,8 @@ const StyledHeadline = styled.div`
 	margin-top: 0;
 	font-family: ${fonts().NORMAL_FONT};
 	font-weight: 500;
+	user-select: none;
```
marionebl: :/ Sure?

src/components/copy/index.tsx
tilmx: I tried to imitate the native behaviour of an native App with that... Do you think it’s not a good idea?

src/components/add-button/index.tsx
tilmx: Just out of curiousity: Why is it better here to split the prop interfaces?

src/components/copy/index.tsx
marionebl: Its the discussion from #519 again, si? 

Personally I am annoyed when apps prevent me from copying things. E.g. I might want to explain what part of the UI I am writing about by citing the text I found there. Being able to copy such text makes my life as a user considerably easier then.

marionebl: That being said there are legit use cases for `user-select: none`, e.g. draggable ui elements

src/components/add-button/index.tsx
tilmx: I removed the text property and instead set the text as a children now.

tilmx:
@marionebl Thanks! I addressed all the issues.

PR #608, opened 2018-09-15
fix: prevent display of highlight area with height/width 0

marionebl:
Address a minor bug where `PreviewStore.highlightArea` would display with a width and height of 0, rendering as 2x2 px rect in the top left corner of the preview.


tilmx:
Works fine for me, thanks!

PR #641, opened 2018-10-29
fix: devtools dependencies needed at runtime

rsteube:
Moved `devtron` and `electron-devtools-installer` to runtime dependencies.


marionebl:
Thank you!

PR #655, opened 2018-11-13
fix: fix minor bugs

### LINKED ISSUE #652: Nested Flex boxes
**Describe the bug**
Not sure if bug, but definetly unexpected: If a flex box component is nested inside a flex box component, the width of the nested component is always 100% what ever is entered (Same for height, if flex-direction is set to vertical)

**To Reproduce**
1. Add a flex box, give it width and height
2. Add another flex box and nest this inside
3. Change width

bildschirmfoto 2018-11-09 um 11 37 49

**Expected behavior**
I want to change width and height properties of nested containers within a flex box container.

tilmx:
## Proposed Changes
- Built-In Box Component: Set flex-grow default value to 0. 
- Remove Page Properties so users aren't irritated by them
- Adjust Built-In Conditional Component texts
- Add "tooltip" to Connect Variable Link Icon
- Do not enter Page Element Name Edit Mode when hitting Enter key.

### Linked Issues
Fixes #652


src/model/pattern-library/builtins/page.ts
```diff
@@ -65,7 +17,7 @@ export const Page = (context: BuiltInContext): BuiltInResult => {
 			id: patternId,
 			name: 'Page',
 			origin: Types.PatternOrigin.BuiltIn,
-			propertyIds: properties.map(p => p.getId()),
+			propertyIds: properties,
```
marionebl: Should be `propertyIds: []`

marionebl:
Deployed at: https://c5p7cr9dc9t62vk3d0ppcubr5xppjvkfe8pp4xb7ctmqgtbk.surge.sh

src/model/pattern-library/builtins/page.ts
tilmx: Thanks, fixed!

PR #668, opened 2018-12-04
perf: put our bundles on a diet

marionebl:
12 MB => 1,5MB

* Code splitting via `react-loadable`
* Enable optimizations in production
* Some optimizations


tilmx:
Already done in #669

PR #669, opened 2018-12-05
feat: edit property with unknown type

marionebl:
* All properties without successful detection are now deemed "unknown"
* Alva renders a "Code properties" detail element if such properties are found
* Input method are small auto-resizing Monaco instances

# Test

* Download attached file: [unknown-properties.alva.zip](https://github.com/meetalva/alva/files/2650494/unknown-properties.alva.zip)
* Open with deployed version: https://alva-issues-669.surge.sh
* Select button 
* See changed properties pane


marionebl:
Deployed at: https://alva-issues-669.surge.sh

tilmx:
Looks good! I'll add some style adjustments. Can we show something like a waiting indicator until the monaco interface is loaded? on my device this takes 2-3 seconds.

PR #679, opened 2018-12-11
fix: get and execute element actions

Palmaswell:
## Proposed Changes
* `elementActions` can either receive a string containing the **action id** or an array containing several **action ids** to be executed.

### Test Case
1. Open a new Alva file
2. Connect the design kit
3. Drag the Input component into the element list
4. Click on the element list name
5. On the right sidebar choose **Handle Change**
5.1 Perform **Set Variable**
5.2 Give a name to your variable e.g. `da bomb` in the named field
5.3 Select **Payload > Value** in the **to** input field
5.4 Select `da bomb` on the **Value** field
6. Type some text in the component

#### Test File
[element-action-test.alva.zip](https://github.com/meetalva/alva/files/2667087/element-action-test.alva.zip)

### DIFF
--- a/src/preview/preview-store.ts
+++ b/src/preview/preview-store.ts
@@ -140,7 +140,6 @@ export class PreviewStore<V> {
 			if (!patternProperty) {
 				return renderProperties;
 			}
-
 			if (patternProperty.getType() === Types.PatternPropertyType.EventHandler) {
 				// Special case:
 				// the link property should render "click" event handlers as href, too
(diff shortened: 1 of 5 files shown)


tilmx:
Looks good to me! I think we should change the commit to 'fix: [...]' instead of 'fix(preview-store): [...]', shouldn't we?

marionebl:
Deployed at: https://alva-issues-679.surge.sh

src/preview/preview-store.ts
```diff
@@ -193,19 +192,23 @@ export class PreviewStore<V> {
 							return;
 						}
 					}
-
-					const elementAction = this.project.getElementActionById(
-						elementProperty.getValue() as string
-					);
-
-					if (!elementAction) {
+					const elementPropertyValues = elementProperty.getValue();
```
marionebl:
```suggestion
                                        const eventProp = elementProperty as Model.PatternEventProperty: // pseudo code
					const rawActionIds = elementProperty.getValue();
					const actionIds = Array.isArray(rawActionIds) ? rawActionIds : [rawActionIds];
					const elementActions = actionIds.map(actionId => this.project.getElementActionById(action)).filter((action): action is Model.ElementAction =>  typeof action !== 'undefined');
					
```

Issue #269, opened 2018-03-08
Skip version update


TheReincarnator:

tilmx:
Maybe this features should be more like a "Skip this version" if I decide this update isn’t worth installing instead of entirely disabling auto-update?

LKuechler:
Currently it is possible to ignore a update for until the app is restarted or the user manually lets Alva check for a update. Is that what should have been done with this issue?

tilmx:
I meant it in a way that the version should be ignored all the time. So like Sketch e.g.:
<img width="732" alt="sketch update" src="https://user-images.githubusercontent.com/8946766/38916858-2c24059e-42e9-11e8-8842-7d902da20a97.png">

tilmx:
Done in latest canary release

PR #701, opened 2018-12-25
feat: enable automatic updates

marionebl:
Reenable automatic updates. Testing is rather involved, e.g. requires a Code Signing Certificate. Detailed instructions here:

https://meetalva.io/doc/docs/howtos/test-autoupdater.html?guides-enabled=true&eo4mrzjvyh-enabled=true

Related: https://github.com/meetalva/alva/issues/269

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,11 @@
     "type": "git",
     "url": "git+https://github.com/meetalva/alva.git"
   },
-  "author": "",
+  "author": {
+    "email": "hey@meetalva.io",
+    "name": "Meet Alva Team",
+    "url": "https://meetalva.io/"
+  },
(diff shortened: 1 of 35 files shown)


marionebl:
Deployed at: https://alva-issues-701.surge.sh

tilmx:
Hm, before merging this PR we should start an AWS machine for the endpoint, right? localhost:9000 is obviously only valid if running minio locally, right?

marionebl:
dev-app-update.yaml only is used in dev Mode, so no real endpoint necessary

tilmx:
Looks good! Let's open a window after a successfull update, so you see when the update is finished.

PR #702, opened 2018-12-25
chore: remove enum options when missing from updated property

marionebl:
## Tests

Status: ✅ @tilmx   

<details>

1. Open Alva
2. Create new file
3. Connect the Design Kit
5. Start designkit build via `tsc -w`
6. Remove `Primary` enum member from `ButtonOrder`
7. Update the pattern library in Alva
8. Add a `Button` to the Element List
9. Only the `Secondary` option should be visible in the `order` select
</details>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,8 +16,8 @@ commands:
 
       - restore_cache:
           keys:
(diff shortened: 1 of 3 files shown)


marionebl:
Deployed at: https://alva-issues-702.surge.sh

PR #703, opened 2018-12-26
WIP: feat: add sentry error tracking

marionebl:
## Proposed Changes

Add sentry.io error tracking so we can gather data about issues of recent releases and unknown bugs to search for based on recurring exceptions.

### ToDo
- [ ] Find out if there is an team OSS plan for sentry.io
- [ ] Check GDPR concerns
- [ ] Decide if we need to let users opt in/out of this


marionebl:
Deployed at: https://alva-site-issues-703.surge.sh

PR #705, opened 2018-12-26
fix: do not coerce optional prop values to empty strings

marionebl:
## Test

Status: ❓   

<details>

1. Open Alva
2. Create new file
3. Connect the internal "circles" library
4. Add a "Copy" element
5. Add a "Text" element as child
6. No error should be thrown

</details>


marionebl:
Deployed at: https://alva-issues-705.surge.sh

PR #706, opened 2018-12-26
fix: simplify app menu mgmt

marionebl:
Fixes a bug where Electron keybindings would not work sporadically.


marionebl:
Deployed at: https://alva-issues-706.surge.sh

PR #709, opened 2018-12-28
fix: allow dragging below (vs. before)

marionebl:


marionebl:
Deployed at: https://alva-issues-709.surge.sh

PR #711, opened 2019-01-02
fix: remove build publishing

tilmx:
... so CircleCI is happy again.


marionebl:
Deployed at: https://alva-issues-711.surge.sh

PR #713, opened 2019-01-02
feat: better splashscreen

marionebl:
* [x] Make it reactive


marionebl:
Deployed at: https://alva-issues-713.surge.sh

PR #714, opened 2019-01-03
fix: update splash screen

tilmx:
Lets add a hover state and restyle the items


marionebl:
Deployed at: https://alva-issues-714.surge.sh

PR #715, opened 2019-01-04
fix: determine if file is saved in app data path correctly on win32

marionebl:
Fixes a bug where the draft state of project files was not computed correctly.


marionebl:
Deployed at: https://alva-issues-715.surge.sh

PR #719, opened 2019-01-05
chore: hello 2019

tilmx:
* update everything to 2019


marionebl:
Deployed at: https://alva-issues-719.surge.sh

tilmx:
> How about we drop the year instead, like e.g. React did: https://github.com/facebook/react/blob/master/LICENSE#L3

Sounds good. I adjusted it to 2017-present.

PR #720, opened 2019-01-05
perf: cut down on app weight by using ncc

marionebl:
* Save weight in final apps by bundling main process stuff via ncc
* Build via `ncc` to `nccbuild`
* Operate `electron-builder`in `nccbuild`
* Introduce `package.ncc.json`, gets copied to `nccbuild`


marionebl:
Deployed at: https://alva-issues-720.surge.sh

PR #727, opened 2019-01-08
perf: avoid repeated deep observer setup

marionebl:


marionebl:
Deployed at: https://alva-issues-727.surge.sh

PR #731, opened 2019-01-11
fix: do not duplicate components with multiple exports

marionebl:
This fixes a bug where importing components with multiple exports would cause duplicated and functionally equivalent entries in the component list.


marionebl:
Deployed at: https://alva-issues-731.surge.sh

marionebl:
Deployed at: https://alva-site-issues-731.surge.sh

PR #732, opened 2019-01-14
perf: switch to message-based edit history

marionebl:
Switch to a message-based edit history in order to avoid huge snapshots.


marionebl:
Deployed at: https://alva-issues-732.surge.sh

marionebl:
Deployed at: https://alva-site-issues-732.surge.sh

tilmx:
Nice work! Here's a list of bugs I came across:

- [ ] Add 3 Text Components. Delete the mid one. Press Cmd+Z. No it's getting weird: You cannot delete the mid one any more.

(to be completed...)

PR #742, opened 2019-01-31
fix: allow components to pass properties

tilmx:


marionebl:
Deployed at: https://alva-site-issues-742.surge.sh

PR #743, opened 2019-01-31
fix: avoid excessive npm data fetching

marionebl:
Ensure we do not needlessly fetch data from NPM on every render.


marionebl:
Deployed at: https://alva-issues-743.surge.sh

marionebl:
Deployed at: https://alva-site-issues-743.surge.sh

PR #746, opened 2019-02-01
fix: restore default values

marionebl:


marionebl:
Deployed at: https://alva-issues-746.surge.sh

PR #748, opened 2019-02-03
feat: detect slot default contents

marionebl:
This change allows users to specify default slot contents. E.g.: 

```ts
export interface ConditionalProps {
  /**
   * @name If False
   *
   * @default
   * ```tsx
   * import * as React from 'react';
   * import { Box, AlignItems, JustifyContent, Text } from '@meetalva/essentials';
   *
   * export default () => (
   *  <Box
   *    flex={true}
   *    justifyContent={JustifyContent.center}
   *    alignItems={AlignItems.center}
   *    height="100px"
   *    backgroundColor="#eee">
   *    <Text text="Shown if condition is false"/>
   *   </Box>
   * );
   * ```
   **/
  falsy?: React.ReactNode;
}
```

will cause Alva to render a `<Box><Text/></Box>` combination for new instances of React components that use `ConditionalProps`.

* [x] Analyze first level of JSX default export in `@default` for props detected as `@slot`
* [x] Determine context id of patterns used in JSX
* [x] Analyze attributes in JSX and expose as "props"
* [x] Escape "@" characters in JSDoc (e.g. import * from '@meetalva/essentials')
* [x] Attach default content candidate as default value to slots
* [x] Translate default contents in `@meetalva/model` to `Model.Element[]` trees
* [x] Normalize JSX Attribute expressions and strings, e.g. `height="100%"` and `height={100}`
* [x] Create Elements for default slot
* [x] Support multiple levels of JSX
* [x] Double check relative paths


marionebl:
Deployed at: https://alva-site-issues-748.surge.sh

PR #762, opened 2019-03-07
feat: add button group property

tilmx:
Adds an option for setting ButtonGroups when using enum properties.

Example annotation:
```
export interface DemoProps {
  /** @control ButtonGroup */
  enumeration?: EnumExample;
}

export enum EnumExample {
  /** @name HALLOWELT @icon Plus */
  Left,

  /** @icon Youtube */
  Center,

  /** @icon Hello */
  Right
}
```

Bildschirmfoto 2019-03-08 um 22 50 18


marionebl:
Deployed at: https://alva-site-issues-762.surge.sh

marionebl:
Deployed at: https://alva-issues-762.surge.sh

packages/components/src/icons/demo.tsx
```diff
@@ -1,51 +1,13 @@
...
-);
-
-const IconRegistryDemo: React.StatelessComponent<void> = (): JSX.Element => (
-	<DemoContainer title="Icons">
-		<DemoIcons size={IconSize.XS} names={IconName} />
-		<DemoIcons size={IconSize.S} names={IconName} />
-	</DemoContainer>
+import { Color } from '../colors';
+
+const IconDemo: React.StatelessComponent<void> = (): JSX.Element => (
```
marionebl: should be `const IconDemo: React.SFC = () => (`

packages/components/src/icons/index.tsx
```diff
@@ -1,91 +1,83 @@
...
+	);
+};
 
-interface IconRegistrySymbolProps {
-	id: string;
+export function isIcon(name: string | undefined): boolean {
+	if (!name) {
+		return false;
+	}
+	if (Object.keys(LocalIcons).includes(name)) {
```
marionebl: Use `.hasOwnProperty` to be consistent with the check against feather icons

packages/components/src/icons/demo.tsx
```diff
@@ -1,51 +1,13 @@
...
+	<>
+		<Icon size={IconSize.XS} color={Color.Green} icon="FlexAlignCenter" />
+		{getIcon({ icon: 'Box' })}
```
marionebl: Ideally there should be no difference between local and remote icons:

```
<Icon icon="Button">
```

should work

marionebl:
I thinks this could be easier by inferring the control type by enums:

1. At least one of the enum members has **no** icon, use `<Select/>`, e.g.
   
   ```ts
   enum E {
       /** @icon A */
       A,
       B
   }   

    export interface D {
      e: E; // E.B has no icon, use select
    }
   ```

2. All of the enum members have an `@icon`, use `<IconGroup/>`

   ```ts
   enum E {
       /** @icon A */
       A,
       /** @icon B */
       B
   }   

    export interface D {
      e: E; // All members have icon, use IconGroup
    }
   ```

tilmx:
> I think this could be easier by inferring the control type by enums:
> 
> 1. At least one of the enum members has **no** icon, use `<Select/>`, e.g.
>    ```ts
>    enum E {
>        /** @icon A */
>        A,
>        B
>    }   
>    
>     export interface D {
>       e: E; // E.B has no icon, use select
>     }
>    ```
> 2. All of the enum members have an `@icon`, use `<IconGroup/>`
>    ```ts
>    enum E {
>        /** @icon A */
>        A,
>        /** @icon B */
>        B
>    }   
>    
>     export interface D {
>       e: E; // All members have icon, use IconGroup
>     }
>    ```

Hm, good idea – but there might be some cases where you want to have a ButtonGroup with Texts (and no Icons), e.g. a Horizontal/Vertical Switch or similar. Or did you mean to implement it optionally (like a fallback)?

marionebl:
> Hm, good idea – but there might be some cases where you want to have a ButtonGroup with Texts (and no Icons), e.g. a Horizontal/Vertical Switch or similar. Or did you mean to implement it optionally (like a fallback)?

What's the added benefit of having a text-only button group compared to a select? My feeling is those text-only groups are really hard to get right in terms of layout and usability, especially with longer labels, etc.

PR #763, opened 2019-03-09
feat: add color picker

marionebl:


packages/core/src/container/property-list/property-item-color.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export interface PropertyItemColorState {
+	propertyOverlay: boolean;
+}
+
+@MobxReact.inject('store')
+@MobxReact.observer
+export class PropertyItemColor extends React.Component<PropertyItemColorProps> {
+	public state = {
```
marionebl: I think we might get away without  internal state in `PropertyItemColor`.

1. `state.color` is a 1:1 copy of `props.property.getValue()`, so let's use that instead
2. `state.displayColorPicker`, `.showPropertyOverlay` and `.hidePropertyOverlay` should be moved to the `PatternProperty`

packages/core/src/container/property-list/property-item-color.tsx
```diff
@@ -0,0 +1,90 @@
...
+	};
+
+	private commit = debounce(() => {
+		const props = this.props as PropertyItemColorProps & { store: ViewStore };
+		props.store.commit();
+	}, 300);
+
+	private handleColorPickerChange = (color: { hex: string }) => {
+		const { property } = this.props;
+		property.setValue(color.hex);
```
marionebl: We might enable the opacity controls by using a string format that supports it - `color.rgb` seems to be available

packages/core/src/container/property-list/property-item-color.tsx
marionebl: Moving `tate.displayColorPicker` to the model would also simplify e.g. handling of clicks on the preview.

marionebl: A good candidate to handle `state.displayColorPicker` might be a `.focused` state on all element properties (not implemented yet). This could control the focus styling on "normal" input fields and the color picker toggling for this.

marionebl:
Screenshot 2019-03-09 at 20 50 17

Minor design inconsistencies

* Right-hand side alignment
* Border color?

marionebl:
Deployed at: https://alva-issues-763.surge.sh

marionebl:
Deployed at: https://alva-site-issues-763.surge.sh

packages/core/src/container/property-list/property-item-color.tsx
yuzl: We had problems with translating the color object from "react-color" (color picker) and "color" (color translation)

PR #767, opened 2019-03-14
fix: ensure correct zIndex for chrome switch

tilmx:
Ensure correct zIndex for Chrome Switch (the switch used for showing/hiding panes on the top left), that you can click on 'Pages' without problems. 

Before that, there was an element with a higher zIndex, so clicking on 'Pages' didn't do anything.

### DIFF
--- a/packages/core/src/container/chrome/chrome-switch.tsx
+++ b/packages/core/src/container/chrome/chrome-switch.tsx
@@ -94,7 +94,9 @@ export class ChromeSwitch extends React.Component {
 									styles={{
 										container: (base: any) => ({
 											...base,
-											display: 'flex'
+											display: 'flex',
+											position: 'relative',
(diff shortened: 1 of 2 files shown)


marionebl:
Deployed at: https://alva-site-issues-767.surge.sh

marionebl:
Deployed at: https://alva-issues-767.surge.sh

packages/core/src/container/chrome/chrome-switch.tsx
```diff
@@ -94,7 +94,9 @@ export class ChromeSwitch extends React.Component {
 									styles={{
 										container: (base: any) => ({
 											...base,
-											display: 'flex'
+											display: 'flex',
+											position: 'relative',
```
yuzl: I don't think `position:relative;` is needed here

Issue #778, opened 2019-04-04
New bug report


levirs565:
Hey there, I just encountered the following error with Alva:

```

```

<details><summary>Stack Trace</summary>

```
Error
    at getBundle (C:\Program Files\Alva Canary\resources\app.asar\index.js:276077:23)
```

</details>

tilmx:
Hey @levirs565, it looks like you tried to connect a library that is not compatible. Can you share the library with us so we could have a look at it?

We are working on a better error management when importing a library, so that it easier to see what didn't work.

levirs565:
I tried to using @meetalva/material-design library from npm

PR #782, opened 2019-04-14
fix: exclude webpack from nccbuild 

marionebl:
#778 meetalva/material-design#2


marionebl:
Deployed at: https://alva-issues-782.surge.sh

marionebl:
Deployed at: https://alva-site-issues-782.surge.sh

